PR #158, opened 2026-04-07
fix: replace mutable default arguments in Config.__init__ with None sentinels

### LINKED ISSUE #130: BUG: Mutable default arguments in Config.__init__ cause shared state
## Description

In `config.py` lines 64-68, default argument values (`StorageDefault()`, `NotifyDefault()`, etc.) are evaluated once at class definition time. All `Config()` instances without explicit arguments share the same provider objects.

## Location

`dotflow/core/config.py:64-68`

```python
def __init__(
    self,
    storage: Storage | None = StorageDefault(),
    notify: Notify | None = NotifyDefault(),
    log: Log | None = LogDefault(),
    api: Api | None = ApiDefault(),
    scheduler: Scheduler | None = SchedulerDefault(),
) -> None:
```

## Impact

If any provider maintains state (e.g., StorageDefault), all Config instances mutate the same shared state. This causes data leak between workflows, especially in parallel/multiprocessing modes.

## Fix

Use `None` as defaults and instantiate inside the method body:

```python
def __init__(self, storage=None, notify=None, ...):
    self.storage = storage or StorageDefault()
    ...
```

## Severity

Critical

avarga1:
Fixes #130

## Problem

Default provider instances in `Config.__init__` were defined as mutable default arguments:

```python
def __init__(
    self,
    storage: Storage | None = StorageDefault(),  # ← evaluated ONCE at class definition
    notify: Notify | None = NotifyDefault(),
    ...
```

In Python, mutable default arguments are evaluated **once** when the class is defined, not on each call. Every `Config()` instance that relies on the defaults shares the same provider objects. Mutating state on one instance's provider silently affects all others.

## Fix

Replace each default with `None` and instantiate fresh providers inside `__init__`:

```python
def __init__(self, storage=None, notify=None, ...):
    self.storage = storage if storage is not None else StorageDefault()
    self.notify = notify if notify is not None else NotifyDefault()
    ...
```

## Test

Added `test_default_providers_are_not_shared_between_instances` which asserts that two `Config()` calls produce distinct provider instances rather than the same shared object.

### DIFF
--- a/dotflow/core/config.py
+++ b/dotflow/core/config.py
@@ -61,17 +61,19 @@ class Config:
 
     def __init__(
         self,
-        storage: Storage | None = StorageDefault(),
-        notify: Notify | None = NotifyDefault(),
-        log: Log | None = LogDefault(),
-        api: Api | None = ApiDefault(),
-        scheduler: Scheduler | None = SchedulerDefault(),
+        storage: Storage | None = None,
+        notify: Notify | None = None,
+        log: Log | None = None,
+        api: Api | None = None,
+        scheduler: Scheduler | None = None,
     ) -> None:
-        self.storage = storage
-        self.notify = notify
-        self.log = log
-        self.api = api
-        self.scheduler = scheduler
+        self.storage = storage if storage is not None else StorageDefault()
+        self.notify = notify if notify is not None else NotifyDefault()
+        self.log = log if log is not None else LogDefault()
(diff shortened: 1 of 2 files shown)


FernandoCelmer:
Hey @avarga1\! Thanks for tackling this one — great catch on the mutable defaults. This was already resolved and merged in PR #150, so I'm closing this to avoid duplication. Really appreciate the contribution, keep them coming\!

PR #159, opened 2026-04-07
fix: ensure retry=0 executes the task exactly once

### LINKED ISSUE #143: BUG: retry=0 causes zero execution attempts, retry=1 is misleading
## Description

Two edge cases in `action.py`:

1. `retry=0` → `range(1, 1)` is empty → function never executes, returns `None` silently
2. `retry=1` (default) → 1 attempt, zero retries. Users setting `retry=1` likely expect 1 retry after the initial attempt.

## Location

`dotflow/core/action.py:84,144`

```python
retry: int = 1,  # default
...
for attempt in range(1, self.retry + 1):
```

## Impact

- `retry=0`: Silent no-op, no error raised
- `retry=1`: Naming confusion — looks like "1 retry" but means "1 total attempt"

## Fix

Either:
- Rename to `max_attempts` to be explicit
- Or guard: `if self.retry < 1: raise ValueError`
- Document clearly that `retry=N` means N total attempts

## Severity

Edge case

avarga1:
## Description

- **dotflow/core/action.py**: Use `max(1, self.retry)` to ensure `retry=0` executes the task at least once
- **tests/core/test_action.py**: Add test verifying `retry=0` runs the task exactly once

## Motivation and Context

With `retry=0`, `range(1, 0+1)` produces an empty range, so the task never executes and returns `None` silently. The fix guarantees at least one execution attempt.

Closes #143

## Types of changes

- [x] Bug fix (change that fixes an issue)
- [ ] New feature (change which adds functionality)
- [ ] Documentation

## Checklist


dotflow/core/action.py
```diff
@@ -141,7 +141,7 @@ def _run_action(self, *args, task=None, **kwargs):
 
         is_async = asyncio.iscoroutinefunction(self.func)
 
-        for attempt in range(1, self.retry + 1):
```
FernandoCelmer: Hey @avarga1! Nice catch on the retry=0 edge case. There's one issue with this approach though.

[Blocking]

**Problem** — The raise guard on line 168 still compares `attempt == self.retry`. With `retry=0`, the loop runs with `attempt=1` (thanks to `max(1, ...)`), but the check becomes `1 == 0` which is `False`. If the task fails, the exception is never re-raised and `_run_action` returns `None` silently.

**Failure scenario** —

```python
@action(retry=0)
def my_task():
    raise ValueError("boom")

# Expected: ValueError raised
# Actual: returns None, task.status = RETRY, sleeps for 1s
```

**Fix** — Extract the computed value and use it in both places:

Replace this:

```python
for attempt in range(1, max(1, self.retry) + 1):
```

With this:

```python
max_attempts = max(1, self.retry)

for attempt in range(1, max_attempts + 1):
```

And on line 168, replace:

```python
if attempt == self.retry:
```

With:

```python
if attempt == max_attempts:
```

This way both the loop bound and the raise guard use the same resolved value.

Also worth adding a test for the failure path:

```python
def test_retry_zero_raises_on_failure(self):
    def always_fail():
        raise ValueError("fail")

    inside = Action(always_fail, retry=0)
    with self.assertRaises(ValueError):
        inside(task=self.task)
```

Everything else looks good!

avarga1:
Good catch — you're right, the raise guard was still comparing against `self.retry` (always `False` when `retry=0` since `attempt` starts at `1`). Fixed by extracting `max_attempts = max(1, self.retry)` and using it in both the loop bound and the raise guard. Also added `test_retry_zero_raises_on_failure` as suggested.

PR #160, opened 2026-04-07
feat: add dotflow viz — terminal pipeline visualization command

### LINKED ISSUE #108: Feature: dotflow viz — terminal pipeline visualization
## Idea

Add a `dotflow viz` CLI command that renders a visual representation of a workflow pipeline in the terminal.

## Motivation

Developers need to visualize their workflow structure — which tasks run, in what order, and with what configuration. Dotflow should provide this visibility without requiring a browser or external service — staying true to the zero-infra, CLI-first philosophy.

### Sequential mode

```
$ dotflow viz workflow.py

  ┌─────────────┐    ┌─────────────┐    ┌─────────────┐
  │ extract_data │───▶│ transform   │───▶│ load_data   │
  │   @action    │    │   @action   │    │   @action   │
  │ retry: 3     │    │ timeout: 30 │    │ retry: 2    │
  └─────────────┘    └─────────────┘    └─────────────┘
```

### Parallel mode

```
$ dotflow viz workflow.py

                     ┌─────────────┐
                 ┌──▶│ process_a   │──┐
  ┌──────────┐   │   └─────────────┘  │   ┌──────────┐
  │ extract  │───┤                    ├──▶│  merge   │
  └──────────┘   │   ┌─────────────┐  │   └──────────┘
                 └──▶│ process_b   │──┘
                     └─────────────┘
```

### After execution (with status)

```
$ dotflow viz workflow.py --result

  ┌─────────────┐    ┌─────────────┐    ┌─────────────┐
  │ extract_data │───▶│ transform   │───▶│ load_data   │
  │  ✅ 0.3s    │    │  ✅ 1.2s    │    │  ❌ 0.1s    │
  │             │    │  retry: 1   │    │  ValueError  │
  └─────────────┘    └─────────────┘    └─────────────┘
```

## Implementation considerations

- Use `rich` library (already a dependency) for terminal rendering
- Parse workflow definition to extract task names, modes, and configuration
- Support both pre-execution (definition) and post-execution (results) views
- Optional `--format mermaid` flag to output Mermaid diagram syntax for docs
- Related to existing issue #16 (Flow diagram dump from code)

## Related issues

- #16 (Flow diagram dump from code)
- #102 (Workflow observability)

avarga1:
Closes #108

## What this adds

A new `dotflow viz` CLI command that renders any workflow as an ASCII pipeline diagram — no browser, no external service, zero new dependencies (uses `rich` which is already in the project).

```
$ dotflow viz -s my_pipeline.workflow

dotflow viz  ·  3 tasks  ·  mode: sequential

┌────────────────┐     ┌────────────────┐     ┌────────────────┐
│ extract        │ ──▶ │ transform      │ ──▶ │ load_data      │
│                │     │                │     │                │
│                │     │ retry: 3       │     │ timeout: 30s   │
└────────────────┘     └────────────────┘     └────────────────┘
```

Parallel mode stacks tasks with a fork bracket:

```
$ dotflow viz -s my_pipeline.workflow -m parallel

dotflow viz  ·  3 tasks  ·  mode: parallel

    ┌────────────────┐
    │ extract        │
  ┌─│                │
  │ └────────────────┘
  │
  ├─┌────────────────┐
  │ │ transform      │
  │ │ retry: 3       │
  │ └────────────────┘
  │
  └─┌────────────────┐
    │ load_data      │
    │ timeout: 30s   │
    └────────────────┘
```

Mermaid export for docs:

```
$ dotflow viz -s my_pipeline.workflow --format mermaid

graph LR
  extract --> transform
  transform --> load_data
```

## Files changed

| File | What |
|---|---|
| `dotflow/utils/visualizer.py` | Core rendering logic — box builder, sequential/parallel/group/mermaid renderers |
| `dotflow/cli/commands/viz.py` | `VizCommand` — wires the CLI args to the visualizer |
| `dotflow/cli/commands/__init__.py` | Exports `VizCommand` |
| `dotflow/cli/setup.py` | Registers `dotflow viz` subcommand with `--step`, `--mode`, `--format` args |
| `tests/utils/test_visualizer.py` | 16 tests covering name resolution, box rendering, sequential, parallel, and mermaid output |

## Usage

```bash
# Sequential (default)
dotflow viz -s my_module.workflow

# Parallel
dotflow viz -s my_module.workflow -m parallel

# Mermaid export
dotflow viz -s my_module.workflow --format mermaid
```


avarga1:
Hey @dotflow-io — just checking in on this one. Happy to address any feedback or rebase if needed. Let me know!

dotflow/utils/visualizer.py
```diff
@@ -0,0 +1,318 @@
...
+        mode:  Execution mode string — 'sequential', 'parallel',
+               'background', or 'sequential_group'.
+        fmt:   Output format — 'terminal' (default) or 'mermaid'.
+    """
+    from dotflow.core.workflow import grouper
+
+    if fmt == "mermaid":
+        print(_render_mermaid(tasks, mode))
+        return
+
```
FernandoCelmer, line 291: [Blocking]

**Problem** — The `else` branch here acts as a catch-all, which means `mode="background"` silently falls through and is labeled as `"sequential"`. The user explicitly requests background mode via `--mode background`, but the header output displays `mode: sequential` — this is actively misleading.

**Failure scenario** — A user runs:
```bash
dotflow flow -s my_module.workflow -m background
```
The output header shows:
```
dotflow flow  ·  3 tasks  ·  mode: sequential
```
The user has no indication that background mode has no distinct rendering, and may incorrectly believe the pipeline will run sequentially.

**Fix** — Either add an explicit branch for background mode, or remove it from the CLI choices:

```python
elif mode == TypeExecution.BACKGROUND:
    diagram = _render_sequential(tasks)
    mode_label = "background"
```

Or, if background has no meaningful distinct visualization, remove `TypeExecution.BACKGROUND` from the `choices` list in `setup_flow()` and document why in a brief comment.

dotflow/utils/visualizer.py
```diff
@@ -0,0 +1,318 @@
...
+    sections = []
+    for group_name, tasks in groups.items():
+        header = f"  ── group: {group_name} ──"
+        body = _render_sequential(tasks)
+        indented = "\n".join("  " + line for line in body.splitlines())
+        sections.append(f"{header}\n{indented}")
+    return "\n\n".join(sections)
+
+
+# ── Mermaid export ────────────────────────────────────────────────────────────
```
FernandoCelmer, line 243: [Suggestion]

**Problem** — Task names are used directly as Mermaid node IDs without deduplication. In Mermaid, node IDs are unique identifiers — if two tasks share the same function name, they collapse into a single node, producing a structurally incorrect diagram.

**Failure scenario** — Consider a pipeline with two tasks both wrapping a function called `process`:

```python
@action
def process():
    pass

wf = DotFlow()
wf.task.add(process)
wf.task.add(process)
```

Running `dotflow flow -s my_module.wf --format mermaid` produces:
```mermaid
graph LR
  process --> process
```
This renders as a **self-loop** on a single node, instead of two distinct nodes connected by an arrow. The user gets a fundamentally wrong picture of their pipeline.

**Fix** — Use a positional suffix as the node ID while keeping the original name as the display label:

```python
def _render_mermaid(tasks: list[Task], mode: str) -> str:
    lines = ["graph LR"]
    names = [_task_name(t) for t in tasks]
    # Deduplicate node IDs to prevent Mermaid collapsing
    node_ids = [f"{name}_{i}" for i, name in enumerate(names)]

    if mode == "parallel":
        lines.append("  START:::hidden")
        lines.append("  END:::hidden")
        for node_id, name in zip(node_ids, names):
            lines.append(f'  START --> {node_id}["{name}"]')
            lines.append(f'  {node_id}["{name}"] --> END')
        lines.append("  classDef hidden display:none")
    else:
        for i in range(len(node_ids) - 1):
            lines.append(f'  {node_ids[i]}["{names[i]}"] --> {node_ids[i+1]}["{names[i+1]}"]')

    return "\n".join(lines)
```

FernandoCelmer (commented):
🔍 Code Review

**Code issues found: 2 blocking/suggestion + 5 inline guidance comments**

| # | Severity | Comment |
|---|----------|---------|
| 1 | [Blocking] | [Background mode displays wrong label](https://github.com/dotflow-io/dotflow/pull/160#discussion_r3068595472) |
| 2 | [Suggestion] | [Duplicate names break Mermaid output](https://github.com/dotflow-io/dotflow/pull/160#discussion_r3068595473) |
| 3 | [Suggestion] | [Remove unnecessary inline comments in viz.py](https://github.com/dotflow-io/dotflow/pull/160#discussion_r3068595475) |
| 4 | [Suggestion] | [Remove narration comments in visualizer.py](https://github.com/dotflow-io/dotflow/pull/160#discussion_r3068595476) |
| 5 | [Suggestion] | [Use rich console instead of bare print()](https://github.com/dotflow-io/dotflow/pull/160#discussion_r3068595478) |
| 6 | [Suggestion] | [Leverage rich Panel/Tree/Columns for rendering](https://github.com/dotflow-io/dotflow/pull/160#discussion_r3068595480) |
| 7 | [Comment] | [Test structure feedback and missing test case](https://github.com/dotflow-io/dotflow/pull/160#discussion_r3068595482) |

dotflow/cli/commands/flow.py
```diff
@@ -0,0 +1,41 @@
...
+
+
+class VizCommand(Command):
+    def setup(self):
+        step = self.params.step
+        mode = self.params.mode
+        fmt = self.params.format
+
+        # Load the workflow step (same module-loading pattern as StartCommand)
+        loaded = Module(value=step)
```
FernandoCelmer, line 14: [Suggestion]

**Problem** — This file has several inline comments that restate what the code already makes obvious. Comments like `# Load the workflow step`, `# DotFlow wraps a TaskBuilder under .task`, `# TaskBuilder exposes .queue directly`, and `# Plain list` add visual noise without providing insight beyond what the code itself communicates. The docstring on `_extract_tasks` already explains the three cases perfectly.

**Principle** — Comments should explain *why*, not *what*. When a comment merely narrates the next line of code, it becomes a maintenance burden (must be kept in sync) without adding understanding. Docstrings, on the other hand, serve as API documentation and should be kept.

**Fix** — Remove the inline comments and rely on the existing docstring:

```python
class FlowCommand(Command):
    def setup(self):
        step = self.params.step
        mode = self.params.mode
        fmt = self.params.format

        loaded = Module(value=step)
        tasks = self._extract_tasks(loaded)
        visualize(tasks=tasks, mode=mode, fmt=fmt)

    @staticmethod
    def _extract_tasks(obj) -> list:
        \"\"\"
        Accept any of the three common ways a user might point us at tasks:
          1. A DotFlow instance          -> obj.task.queue
          2. A TaskBuilder instance      -> obj.queue
          3. A plain list[Task]          -> obj
        \"\"\"
        if hasattr(obj, "task") and hasattr(obj.task, "queue"):
            return list(obj.task.queue)

        if hasattr(obj, "queue"):
            return list(obj.queue)

        if isinstance(obj, list):
            return obj

        return []
```

Keep docstrings — they document intent and contract. Remove comments that only restate the code.

dotflow/utils/visualizer.py
```diff
@@ -0,0 +1,318 @@
...
+_BRANCH_TOP = "┌"
+_BRANCH_BOT = "└"
+
+
+def _task_name(task: Task) -> str:
+    step = task.step
+
+    # Case 1: @action direct — step is an Action instance with .func set
+    func = getattr(step, "func", None)
+    if func is not None:
```
FernandoCelmer, line 35: [Suggestion]

**Problem** — Throughout this file there are inline comments that narrate *what* the code does rather than *why*. This includes:

**Narration comments (explain what the next line does):**
- `# Case 1: @action direct — step is an Action instance with .func set` (line 35)
- `# Case 2: @action(retry=N) — step is the inner closure returned by Action.__call__` (line 40)
- `# The original function is captured as args[0] in the closure.` (line 41)
- `# @action direct -> step is the Action instance` (line 59)
- `# @action(retry=N) -> step is a closure; Action instance is in 'self' cell` (line 61)
- `# subtract border chars` (line 110)
- `# Insert fillers before the last line (bottom border)` (line 141)
- `# always connect on the first content row (row after top border)` (line 148)
- `# Left bracket` (line 195)

**Section dividers (visual noise, add no information):**
- `# ── Box-drawing constants ───────────���───────────────────────────────────────────`
- `# ── Box builder ────────────────────────────────────��──────────────────────────`
- `# ── Sequential renderer ──────────────────────��────────────────────────────────`
- `# ── Parallel renderer ──────────────────────────────────────────────────���──────`
- `# ── Group renderer ─────────────────────────────────────────────────────────���──`
- `# ── Mermaid export ───────────────��────────────────────────────────────────────`
- `# ── Public API ───────────────────���─────────────────────────���──────────────────`

These dividers should also be removed. The code structure is already clear from the function/class names and docstrings — if you need to visually separate sections, use blank lines (which you already do). The dividers are redundant and create maintenance overhead.

**Principle** ��� Code should be self-documenting. Docstrings are great — keep them. But inline comments should only explain *why* something non-obvious is done, never *what* the code does. Section dividers belong in documentation, not in source code — the file structure and function names already provide that navigation.

**Fix** — Remove all narration comments AND section dividers. If the closure introspection logic truly needs explanation, rewrite as a *why* comment:

```python
# @action(retry=N) wraps the original function in a closure.
# We must introspect __closure__ to recover the original function name
# because the wrapper doesn't preserve __name__.
```

Keep all existing docstrings — they serve as API documentation and are appropriate.

dotflow/utils/visualizer.py
```diff
@@ -0,0 +1,318 @@
...
+
+
+def visualize(
+    tasks: list[Task],
+    mode: str = "sequential",
+    fmt: str = "terminal",
+) -> None:
+    """
+    Render a workflow pipeline to the terminal or as Mermaid markup.
+
```
FernandoCelmer, line 279: [Suggestion]

**Problem** — This line uses bare `print()` while the rest of the module already imports and uses `rich.console.Console`. Since `rich` is already a dependency of dotflow (declared in `pyproject.toml`), you should use it consistently for all terminal output. Using `print()` bypasses rich's output handling (width detection, encoding, redirects).

**Fix** — Use the existing `console` instance:

```python
if fmt == "mermaid":
    console.print(_render_mermaid(tasks, mode), highlight=False)
    return
```

The `highlight=False` prevents rich from applying syntax highlighting to the Mermaid markup, keeping it clean for copy-paste.

dotflow/utils/visualizer.py
```diff
@@ -0,0 +1,318 @@
...
+from rich.console import Console
+from rich.text import Text
+
+if TYPE_CHECKING:
+    from dotflow.core.task import Task
+
+console = Console()
+
+# ── Box-drawing constants ─────────────────────────────────────────────────────
```
FernandoCelmer, line 15: [Suggestion]

**Problem** — The module currently builds box-drawing diagrams manually with string concatenation (`_build_box`, `_render_sequential`, `_render_parallel`). While this works, `rich` — which is already imported here and is a declared dotflow dependency — provides high-level components specifically designed for this kind of terminal rendering:

- **`rich.panel.Panel`** — renders bordered boxes with titles, padding, and styles (replaces manual `_build_box`)
- **`rich.columns.Columns`** — lays out renderables side by side (replaces manual sequential arrow logic)
- **`rich.tree.Tree`** — renders tree/hierarchy structures with branch lines (replaces manual parallel bracket logic)
- **`rich.table.Table`** — can render tasks in a structured grid format

**Benefits of using rich components:**
1. Automatic terminal width handling — no hardcoded `width=18`
2. Proper Unicode support and fallback for terminals that don't support box-drawing
3. Style/color support built-in (bold task names, dim metadata, colored status)
4. Responsive layout that adapts to terminal size

**Example** — Sequential rendering with rich:

```python
from rich.panel import Panel
from rich.columns import Columns
from rich.text import Text

def _render_sequential_rich(tasks: list[Task]) -> Columns:
    panels = []
    for i, task in enumerate(tasks):
        name = _task_name(task)
        body = "\n".join(_task_config(task))
        panel = Panel(body or " ", title=name, width=20, style="cyan")
        panels.append(panel)
        if i < len(tasks) - 1:
            panels.append(Text(" -->  ", style="bold"))
    return Columns(panels, equal=False, expand=False)
```

**Example** — Parallel rendering with rich Tree:

```python
from rich.tree import Tree
from rich.panel import Panel

def _render_parallel_rich(tasks: list[Task]) -> Tree:
    tree = Tree("parallel", style="bold yellow")
    for task in tasks:
        name = _task_name(task)
        body = "\n".join(_task_config(task))
        tree.add(Panel(body or " ", title=name, width=20))
    return tree
```

This is not a blocking issue — the current implementation works correctly. But leveraging rich's built-in components would reduce code complexity, improve visual output, and make future enhancements (colors, styles, responsive width) trivial.

tests/utils/test_visualizer.py
```diff
@@ -0,0 +1,132 @@
+"""Tests for the workflow visualizer."""
+
```
FernandoCelmer, line 2: [Comment]

The tests are well-structured and cover the key rendering paths — good coverage of name resolution, box building, sequential, parallel, and mermaid modes. A few notes:

1. **Keep the docstring** at the top (`"Tests for the workflow visualizer."`), it's appropriate module-level documentation.

2. **Test methods are self-explanatory** — their names (`test_direct_decorator`, `test_box_has_top_and_bottom_border`, etc.) document intent perfectly. No need to add inline comments here either. This is a great example of self-documenting test code.

3. **Consider adding a test for duplicate task names in Mermaid output** — given the issue flagged above where duplicate function names collapse into a single Mermaid node, a test like this would catch the regression:

```python
def test_mermaid_duplicate_names_produce_distinct_nodes(self):
    tasks = _make_tasks(step_a, step_a)
    output = _render_mermaid(tasks, mode="sequential")
    # Should have two distinct nodes, not a self-loop
    lines = [l.strip() for l in output.splitlines() if "-->" in l]
    self.assertEqual(len(lines), 1)
    # The two sides of the arrow should be different node IDs
    parts = lines[0].split("-->")
    self.assertNotEqual(parts[0].strip(), parts[1].strip())
```

Great work on the test structure overall!

FernandoCelmer:
Hey! 👋

Just a heads-up — this branch currently has merge conflicts with `develop`. Could you rebase or merge `develop` into your branch when you get a chance? That way we can get this reviewed and merged smoothly.

```bash
git checkout feature/ISSUE-108
git fetch origin
git merge origin/develop
# resolve conflicts
git push
```

---

💡 **Tip:** If you're using [Claude Code](https://docs.anthropic.com/en/docs/claude-code), there's a git-flow skill that automates branch naming, commits, and rebasing following the project conventions:

👉 https://github.com/FernandoCelmer/skills/blob/master/skills/git-flow/SKILL.md

You can install it with:
```bash
claude plugins install git-flow@FernandoCelmer-skills
```

It handles the `feature/ISSUE-NUMBER` branch pattern and conventional commit messages automatically — super handy for keeping everything consistent!

FernandoCelmer:
Hey! One more thing — could you rename the command from `dotflow viz` to `dotflow flow`? This applies to:

- The CLI subcommand registration (`add_parser("viz", ...)` → `add_parser("flow", ...)`)
- The class name (`VizCommand` → `FlowCommand`)
- The file name (`commands/viz.py` → `commands/flow.py`)
- The setup method (`setup_viz()` → `setup_flow()`)
- The `__all__` export and imports
- Help texts and docstrings referencing "viz"

`dotflow flow` reads more naturally as a pipeline visualization command and aligns better with the project's naming conventions.

dotflow/utils/visualizer.py
FernandoCelmer: Thanks for fixing this! The explicit `elif mode == "background"` branch is exactly what was needed. Resolved! 🎉

dotflow/utils/visualizer.py
FernandoCelmer: Great implementation with `node_ids` and positional suffixes — duplicate names are now handled correctly. The new test covers it too. Resolved! 🎉

dotflow/cli/commands/flow.py
FernandoCelmer: Clean! No more narration comments, docstring preserved. Renaming to `flow.py` and `FlowCommand` looks great. Resolved! 🎉

dotflow/utils/visualizer.py
FernandoCelmer: Nice work removing the narration comments and rewriting the closure introspection comment to explain *why*. The section dividers (`# ── Box-drawing constants ──`, `# ── Box builder ──`, etc.) are still present — consider removing them in a follow-up since the file structure is already clear from function names and docstrings. Resolving this one! 🎉

dotflow/utils/visualizer.py
FernandoCelmer: Perfect — `console.print(..., highlight=False)` is the right call. Consistent rich usage throughout. Resolved! 🎉

dotflow/utils/visualizer.py
FernandoCelmer: This was a suggestion for a future improvement — the current implementation works well. Feel free to explore `rich.panel.Panel` and `rich.tree.Tree` in a follow-up if you'd like to enhance the visual output. Resolving! 🎉

tests/utils/test_visualizer.py
FernandoCelmer: The new `test_duplicate_names_produce_distinct_nodes` test is a great addition — validates the Mermaid deduplication fix. Test suite looks solid. Resolved! 🎉

PR #162, opened 2026-04-07
fix: use regex to match method definitions in execution orderer

### LINKED ISSUE #134: BUG: Method ordering uses string search that can match wrong methods
## Description

In `execution.py` line 78, method ordering uses `code.find(f"def {callable_name}")` to determine source position. This can match comments, strings, or methods with similar prefixes (e.g., `run` matches `run_all`).

## Location

`dotflow/core/execution.py:78`

## Impact

Methods may be executed in wrong order if names are prefixes of each other or appear in comments.

## Fix

Use `inspect.getsourcelines()` per method, or match with a regex like `r"def {name}\s*\("`.

## Severity

Significant

avarga1:
## Summary

Fixes #134

`_execution_orderer` in `execution.py` used `str.find(f"def {callable_name}")` to locate each method in the class source. This matches any line containing that substring, so a method named `run` would incorrectly match `def run_all(...)`, causing wrong execution order or duplicate entries.

**Root cause** (`execution.py` line 78, before fix):
```python
if code.find(f"def {callable_name}") != -1:
```

**Fix:** compile a regex that requires the name to be followed by `\s*(` — i.e., only a true `def run(...)` declaration matches:
```python
pattern = re.compile(rf"def {re.escape(callable_name)}\s*\(")
if pattern.search(code):
```

`re.escape` also guards against method names containing regex metacharacters.

## Changes

- `dotflow/core/execution.py`: import `re`; replace `str.find` with `re.compile(...).search`
- `tests/mocks/step_class.py`: add `ActionStepPrefixMethods` (methods `run` and `run_all`)
- `tests/mocks/__init__.py`: export `ActionStepPrefixMethods`
- `tests/core/test_execution.py`: add `test_execution_orderer_prefix_methods` — verifies each method appears exactly once and `run` precedes `run_all`

## Test plan

- [x] All 16 `tests/core/test_execution.py` tests pass
- [x] New test `test_execution_orderer_prefix_methods` explicitly catches the prefix-match regression

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/dotflow/core/execution.py
+++ b/dotflow/core/execution.py
@@ -1,5 +1,6 @@
 """Execution module"""
 
+import re
 from collections.abc import Callable
 from datetime import datetime
 from inspect import getsourcelines
@@ -74,8 +75,9 @@ class Execution:
             inside_code = getsourcelines(class_instance.__class__)[0]
 
             for callable_name in callable_list:
+                pattern = re.compile(rf"def {re.escape(callable_name)}\s*\(")
                 for index, code in enumerate(inside_code):
-                    if code.find(f"def {callable_name}") != -1:
+                    if pattern.search(code):
                         ordered_list.append((index, callable_name))
 
             ordered_list.sort()
--- a/tests/core/test_execution.py
+++ b/tests/core/test_execution.py
@@ -13,6 +13,7 @@ from dotflow.core.types import TypeStatus
 from tests.mocks import (
     ActionStep,
     ActionStepExecutionOrderer,
+    ActionStepPrefixMethods,
     ActionStepWithContexts,
     ActionStepWithError,
     ActionStepWithInitialContext,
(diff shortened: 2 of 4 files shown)


FernandoCelmer:
Hey @avarga1! Thanks for this — really solid work, especially the test coverage with `ActionStepPrefixMethods`.

I already pushed a fix for this in PR #163 with a slightly different approach:

- Uses `\b` word boundary for stricter matching
- Uses `\s+` instead of a literal space between `def` and the method name
- Adds a `break` after the first match to avoid duplicate entries in the ordered list

Since #163 is already merged, I'm going to close this one. But the test you wrote is great — would you be open to opening a separate PR just with the test (`ActionStepPrefixMethods` mock + `test_execution_orderer_prefix_methods`)? It would be a valuable addition to the test suite.

Thanks again for the contribution!

PR #164, opened 2026-04-07
test: add regression test for method-name prefix matching in _execution_orderer

avarga1:
## Summary

As requested in #162 — this PR adds only the test coverage for the bug fixed in #163, with no changes to production code.

## What's added

- `tests/mocks/step_class.py`: `ActionStepPrefixMethods` — a class with methods `run` and `run_all` where one name is a prefix of the other
- `tests/mocks/__init__.py`: exports `ActionStepPrefixMethods`
- `tests/core/test_execution.py`: `test_execution_orderer_prefix_methods` — asserts that `run` and `run_all` each appear exactly once in the result, and that `run` precedes `run_all`

## Note on CI

This test **intentionally fails on master** without the fix from #163 (the old `str.find()` matches `run` against `def run_all(...)`, producing a duplicate entry). It passes once #163 is merged.

Happy to rebase onto `feature/134` if that's easier to merge together.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/tests/core/test_execution.py
+++ b/tests/core/test_execution.py
@@ -13,6 +13,7 @@ from dotflow.core.types import TypeStatus
 from tests.mocks import (
     ActionStep,
     ActionStepExecutionOrderer,
+    ActionStepPrefixMethods,
     ActionStepWithContexts,
     ActionStepWithError,
     ActionStepWithInitialContext,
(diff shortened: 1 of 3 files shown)


avarga1:
Retargeted to `feature/134` so the test runs against your fix and CI passes. The failure on `master` was expected — that's exactly the regression the test catches.

Issue #201, opened 2026-04-08
Refactor: Separate Engine from Execution Strategy


FernandoCelmer:
## Motivation

Currently, the **task lifecycle** logic (managing status, handling errors, measuring duration) is mixed with the logic of **how tasks are executed** (sequential, parallel, background). This creates two problems:

1. **`Execution` does too much** — it runs in `__init__`, mixing status control, function invocation, error handling, and callbacks into a single `_execution()` method.
2. **Strategies (`Sequential`, `Parallel`, etc.) duplicate logic** — each one re-implements checkpoint handling, `previous_context` control, and the call to `Execution`.

Separating these responsibilities makes the code more testable, extensible, and easier to understand.

---

## Proposal

Introduce a **`TaskEngine`** that encapsulates the task lifecycle (status, retry, timeout, error handling, duration), and keep the execution strategies (`Sequential`, `Parallel`, etc.) responsible only for **ordering and parallelism**.

### Proposed Architecture

```
Strategy (Sequential, Parallel, ...)
  └── decides the ORDER of execution
       └── TaskEngine
            └── manages the LIFECYCLE of each task
                 └── calls the user function (@action)
```

---

### 1. `Execution.__init__` runs the task in the constructor

```python
# dotflow/core/execution.py (current)
class Execution:
    def __init__(self, task, workflow_id, previous_context=None, _flow_callback=basic_callback):
        self.task = task
        self.task.status = TypeStatus.IN_PROGRESS
        self.task.previous_context = previous_context
        self.task.workflow_id = workflow_id
        self._execution(_flow_callback)  # ← runs in __init__!
```

This prevents any external control over the execution lifecycle. For example, it's not possible to pause between creation and execution, or inject intermediate behavior.

### 2. `_execution()` mixes responsibilities

```python
# dotflow/core/execution.py (current)
def _execution(self, _flow_callback):
    try:
        start = datetime.now()
        current_context = self.task.step(...)        # calls the function
        # ... handles class result ...
        self.task.current_context = current_context   # persists
        self.task.duration = (datetime.now() - start).total_seconds()  # measures time
        self.task.status = TypeStatus.COMPLETED       # changes status
    except Exception as err:
        self.task.errors = err                        # records error
        self.task.status = TypeStatus.FAILED          # changes status
    finally:
        self.task.callback(task=self.task)             # task callback
        _flow_callback(task=self.task)                 # flow callback
```

[...]

Issue #131, opened 2026-04-06
BUG: Timeout via ThreadPoolExecutor does not kill the running thread


FernandoCelmer:
## Description

In `action.py` lines 147-154, when `future.result(timeout=self.timeout)` raises `TimeoutError`, the underlying thread continues running in the background. Python's `ThreadPoolExecutor` cannot cancel running threads.

## Location

`dotflow/core/action.py:147-154`

```python
if self.timeout:
    with ThreadPoolExecutor(max_workers=1) as executor:
        future = executor.submit(self._call_func, is_async, *args, **kwargs)
        result = future.result(timeout=self.timeout)
```

## Impact

Resource leak — timed-out functions keep running, consuming CPU/memory and potentially causing side effects.

## Severity

Significant

FernandoCelmer:
Closing — the current implementation in `engine.py` already applies the best mitigation available with Python threads:

```python
future.cancel()
executor.shutdown(wait=False, cancel_futures=True)
```

The fundamental limitation is that **Python cannot kill a running thread**. `future.cancel()` only works if the future hasn't started, and `shutdown(cancel_futures=True)` cancels queued futures, not threads already executing.

This is a known Python limitation, not a Dotflow bug. The current implementation is the best that can be done with `ThreadPoolExecutor`.

A separate issue will be opened for a future enhancement using `multiprocessing` to achieve real timeout enforcement via `process.terminate()`.

PR #202, opened 2026-04-08
⚙️ FEATURE-#201: Separate Engine from Execution Strategy

FernandoCelmer:
# Description

Refactor the task execution architecture by introducing a `TaskEngine` that encapsulates the task lifecycle, separating it from execution strategies. Moves retry, timeout and backoff from `@action` to the engine.

Issue: [📌 ISSUE-#201](https://github.com/dotflow-io/dotflow/issues/201)

### Phase 1 — TaskEngine
- **`TaskEngine`** (`dotflow/core/engine.py`) — manages lifecycle via context manager: status transitions, duration, tracer, error handling
- **`Execution`** — refactored as backward-compatible wrapper around `TaskEngine`
- **`Flow` base class** — `_has_checkpoint()` and `_restore_checkpoint()` extracted (DRY)
- **Strategies** (`Sequential`, `SequentialGroup`, `Background`) — use `TaskEngine` directly

### Phase 2 — Retry/Timeout in Engine
- **`execute_with_retry()`** — retry loop, timeout via `ThreadPoolExecutor`, backoff
- **`Action._run_action()`** — simplified to single attempt
- **`@action` attributes** — `retry`, `timeout`, `retry_delay`, `backoff` propagated to closure

### Phase 3 — Composable Context Managers
- **`checkpoint_context()`** — automatic checkpoint post-execution
- Docs updated to reference `TaskEngine`

### Bug Fix
- Checkpoint example now uses valid UUID

## Type of change

- [x] New feature (non-breaking change which adds functionality)

## Validation

- 427 tests passing (21 new for TaskEngine)
- All `docs_src/` examples verified
- flake8, ruff check, ruff format — all clean

# Checklist:

- [x] New and existing unit tests pass locally with my changes
- [x] I have made corresponding changes to the documentation


dotflow/core/engine.py
```diff
@@ -0,0 +1,277 @@
...
+
+        return current_context
+
+    def _execute_with_timeout(self, seconds: int):
+        """Executes the task function with a real timeout using ThreadPoolExecutor."""
+        executor = ThreadPoolExecutor(max_workers=1)
+        try:
+            future = executor.submit(self._execute_single)
+            return future.result(timeout=seconds)
+        except TimeoutError:
```
FernandoCelmer, line 173: [Blocking]

**Problem** — The `ThreadPoolExecutor` created in `_execute_with_timeout()` **is not shut down on the success path**. The `except TimeoutError` and `except Exception` blocks call `executor.shutdown()`, but when `future.result()` returns normally, the function does `return` without ever calling `shutdown()`. This re-introduces the resource leak that was fixed in BUG-#131.

In the old `action.py` code, there was an explicit `else: executor.shutdown(wait=False)` block that was dropped during the migration to `TaskEngine`.

**Failure scenario** — Every task with `timeout > 0` that completes successfully leaks an executor thread. In a workflow with hundreds of timeout-enabled tasks, the thread accumulation eventually causes `RuntimeError: can't start new thread`, degrading or crashing the system in production.

**Fix** — Add an `else` block or use `finally` to guarantee shutdown:

```python
def _execute_with_timeout(self, seconds: int):
    executor = ThreadPoolExecutor(max_workers=1)
    try:
        future = executor.submit(self._execute_single)
        return future.result(timeout=seconds)
    except TimeoutError:
        future.cancel()
        executor.shutdown(wait=False, cancel_futures=True)
        raise
    except Exception:
        executor.shutdown(wait=False)
        raise
    else:
        executor.shutdown(wait=False)
```

Alternatively, use `try/finally` to cover all paths:

```python
def _execute_with_timeout(self, seconds: int):
    executor = ThreadPoolExecutor(max_workers=1)
    try:
        future = executor.submit(self._execute_single)
        return future.result(timeout=seconds)
    except TimeoutError:
        future.cancel()
        raise
    finally:
        executor.shutdown(wait=False, cancel_futures=True)
```

dotflow/core/engine.py
```diff
@@ -0,0 +1,277 @@
...
+        try:
+            yield self
+        except AssertionError as err:
+            raise err
+        except Exception as err:
+            self.task.errors = err
+            self.task.current_context = None
+            self.task.status = TypeStatus.FAILED
+        else:
+            self.task.duration = (
```
FernandoCelmer, line 81: [Blocking]

**Problem** — `task.duration` is calculated in the `else` block of the `start()` context manager (success path only), but `tracer.end_task()` is called in the `finally` block. This means that **on any failure**, the tracer receives `task.duration = None` because the `else` block never executes when an exception is raised.

The old code in `execution.py` calculated duration **before** calling the tracer, ensuring the value was always available.

**Failure scenario** — Every task that fails (exception, timeout, validation error) reports `duration=None` in OpenTelemetry spans. Monitoring dashboards and duration-based alerts end up with incomplete data — impossible to calculate p95/p99 latency for failed tasks, impossible to detect tasks that are progressively slowing down before they fail.

**Fix** — Move the duration calculation to the `finally` block, before `tracer.end_task()`:

```python
@contextmanager
def start(self):
    self.task.workflow_id = self.workflow_id
    self.task.previous_context = self.previous_context
    self.task.config.tracer.start_task(task=self.task)
    self.task.status = TypeStatus.IN_PROGRESS
    self._start_time = datetime.now()

    try:
        yield self
    except AssertionError as err:
        raise err
    except Exception as err:
        self.task.errors = err
        self.task.current_context = None
        self.task.status = TypeStatus.FAILED
    else:
        if self.task.status in (
            TypeStatus.IN_PROGRESS,
            TypeStatus.RETRY,
        ):
            self.task.status = TypeStatus.COMPLETED
    finally:
        self.task.duration = (
            datetime.now() - self._start_time
        ).total_seconds()
        self.task.config.tracer.end_task(task=self.task)
```

dotflow/core/engine.py
```diff
@@ -0,0 +1,277 @@
...
+            future.cancel()
+            executor.shutdown(wait=False, cancel_futures=True)
+            raise
+        except Exception:
+            executor.shutdown(wait=False)
+            raise
+
+    @staticmethod
+    def _is_class_internal_error(error: Exception) -> bool:
+        """Checks if an error is an internal class execution error."""
```
FernandoCelmer, line 183: [Blocking]

**Problem** — There is a flaw in the interaction between `action.py` and `engine.py` for class-based error detection. The current flow is:

1. `action.py` → `_run_action()` catches a `TypeError` (e.g., `missing 1 required positional argument: 'self'`)
2. `action.py` → converts it to `ExecutionWithClassError()` with message `"Unknown"` via `raise ExecutionWithClassError() from None`
3. `engine.py` → `execute_with_retry()` catches the exception and calls `_is_class_internal_error(error)`
4. `engine.py` → `_is_class_internal_error` looks for `"initial_context"`, `"previous_context"` or `"missing 1 required positional argument: 'self'"` in the message
5. **The message is now `"Unknown"`** (or the default from `ExecutionWithClassError`), so no pattern matches
6. The guard returns `False` and the error is **retried** instead of failing immediately

**Failure scenario** — When an `@action` decorates a class method (not a function), the TypeError from the signature mismatch is caught by `action.py` and converted before `engine.py` can detect it. Result: the retry loop attempts N times an error that will never be resolved by retrying, wasting time (`retry_delay * N`) and producing N confusing error log entries.

**Fix** — Make `_is_class_internal_error` also check for the `ExecutionWithClassError` type:

```python
@staticmethod
def _is_class_internal_error(error: Exception) -> bool:
    if isinstance(error, ExecutionWithClassError):
        return True
    message = str(error)
    patterns = [
        "initial_context",
        "previous_context",
        "missing 1 required positional argument: 'self'",
    ]
    return any(pattern in message for pattern in patterns)
```

dotflow/abc/flow.py
```diff
@@ -40,3 +42,24 @@ def _flow_callback(self, task: Task) -> None:
     @abstractmethod
     def run(self) -> None:
         return None
+
+    def _has_checkpoint(self, task: Task) -> bool:
+        if not self.resume:
+            return False
+
```
FernandoCelmer, line 49: [Suggestion]

**Problem** — The DRY extraction of `_has_checkpoint()` and `_restore_checkpoint()` introduced a double storage read for the same key. `_has_checkpoint()` calls `storage.get()` to check if a checkpoint exists, and immediately after, `_restore_checkpoint()` calls `storage.get()` again with the same key to retrieve the same value.

This pattern repeats in 3 locations: `Sequential.run()`, `SequentialGroup._run_group()`, and `Background._run_sequential()`.

**Failure scenario** — With a networked storage provider (S3, Redis, HTTP), each task resume generates 2 round-trips instead of 1. Resuming 50 tasks: `50ms × 50 tasks × 2 = 5s` of I/O instead of `2.5s`. In workflows with hundreds of checkpointed tasks, the additional latency becomes significant.

**Fix** — Unify the check and restore into a single method, or have `_has_checkpoint` return the `Context`:

```python
def _try_restore_checkpoint(self, task: Task) -> Context | None:
    """Attempts to restore a checkpoint. Returns the Context if found, None otherwise."""
    if not self.resume:
        return None

    context = task.config.storage.get(
        key=task.config.storage.key(task=task)
    )

    if context.storage is None:
        return None

    task.status = TypeStatus.COMPLETED
    task.current_context = context
    self._flow_callback(task=task)

    return context
```

And in the workflows:
```python
restored = self._try_restore_checkpoint(task)
if restored:
    previous_context = restored
    continue
```

dotflow/core/action.py
```diff
@@ -123,77 +120,33 @@ def action(*_args, **_kwargs):
...
+        action.timeout = self.timeout
+        action.retry_delay = self.retry_delay
+        action.backoff = self.backoff
 
-    def _run_action(self, *args, task=None, **kwargs):
-        current_delay = self.retry_delay
+        return action
 
+    def _run_action(self, *args, **kwargs):
         is_async = asyncio.iscoroutinefunction(self.func)
```
FernandoCelmer, line 142: [Suggestion]

**Problem** — The `Action` class docstring (lines 22-75) still documents retry, timeout, backoff, and retry_delay as behaviors of `Action` itself, including examples like `@action(retry=5)` and `@action(timeout=60)`. However, after this refactor, `Action` only **stores** these parameters as attributes — all execution logic has been moved to `TaskEngine.execute_with_retry()`.

The `_run_action()` method is now a single call with no retry loop, but the class docstring still describes the full retry cycle with backoff.

**Failure scenario** — A developer reading the `@action` docstring believes the decorator alone implements retry and timeout. When writing unit tests for `Action` directly (without `TaskEngine`), they cannot reproduce the documented retry behavior. Or worse: they assume `Action` can be called outside `TaskEngine` and still get automatic retry.

**Fix** — Update the docstring to reflect that `Action` is the **configuration point** and `TaskEngine` is the **executor**:

```python
class Action:
    """Decorator for creating task steps.

    Accepts configuration for retry, timeout, retry_delay, and backoff,
    which are stored as attributes and consumed by TaskEngine during execution.

    Usage:
        @action
        def my_task():
            ...

        @action(retry=5, timeout=60)
        def my_task():
            ...

    Note: retry, timeout, and backoff are enforced by TaskEngine,
    not by Action itself. See TaskEngine.execute_with_retry().
    """
```

FernandoCelmer (commented):
🔍 Code Review

**Code issues found: 6**

| # | Severity | Comment |
|---|----------|---------|
| 1 | [Blocking] | [Executor leak on success path](https://github.com/dotflow-io/dotflow/pull/202#discussion_r3054578326) |
| 2 | [Blocking] | [Duration None when tracer calls end_task](https://github.com/dotflow-io/dotflow/pull/202#discussion_r3054578339) |
| 3 | [Blocking] | [ExecutionWithClassError bypasses retry guard](https://github.com/dotflow-io/dotflow/pull/202#discussion_r3054578344) |
| 4 | [Suggestion] | [Double storage read on checkpoint](https://github.com/dotflow-io/dotflow/pull/202#discussion_r3054578347) |
| 5 | [Suggestion] | [Action docstring outdated after refactor](https://github.com/dotflow-io/dotflow/pull/202#discussion_r3054578352) |
| 6 | [Suggestion] | Parallel still uses deprecated Execution (see below) |

---

### 6. [Suggestion] Parallel still uses deprecated Execution

**File**: `dotflow/core/workflow.py` ~line 430

**Problem** — The new `Execution` docstring explicitly states: *"New code should use TaskEngine directly."* The strategies `Sequential`, `SequentialGroup`, and `Background` were migrated to `TaskEngine`, but `Parallel.run()` still uses `Execution`:

```python
process = _mp.Process(
    target=Execution,
    args=(task, self.workflow_id, previous_context, self._flow_callback),
)
```

This creates an architectural inconsistency: 3 out of 4 strategies use the new pattern, 1 uses the old one.

**Failure scenario** — Future developers may use `Parallel` as a reference and perpetuate the use of `Execution` instead of `TaskEngine`. Additionally, any bug fix or improvement made to `TaskEngine` will not automatically apply to tasks executed via `Parallel`, creating behavioral divergence between execution modes.

**Fix** — If the `Parallel` migration was intentionally deferred (e.g., multiprocessing pickling complexity), add an explicit TODO comment:

```python
# TODO: Migrate Parallel to use TaskEngine directly.
# Currently uses Execution wrapper due to multiprocessing pickling constraints.
process = _mp.Process(
    target=Execution,
    args=(task, self.workflow_id, previous_context, self._flow_callback),
)
```

dotflow/core/engine.py
FernandoCelmer: Fixed in PR #222 —  now uses `finally: executor.shutdown(wait=False, cancel_futures=True)` covering all paths.

dotflow/core/engine.py
FernandoCelmer: Fixed in PR #222 — `task.duration` moved to `finally` block, always calculated before `tracer.end_task()`.

dotflow/core/engine.py
FernandoCelmer: Fixed in PR #222 — `_is_class_internal_error` now checks `isinstance(error, ExecutionWithClassError)` before pattern matching.

dotflow/abc/flow.py
FernandoCelmer: Fixed in PR #222 — Unified into `_try_restore_checkpoint()` with a single storage query.

dotflow/core/action.py
FernandoCelmer: Fixed in PR #222 — Action docstring updated to clarify that retry/timeout are enforced by TaskEngine.

Issue #217, opened 2026-04-08
Bug: SerializerTask.model_dump_json() mutates self and produces truncated invalid JSON


FernandoCelmer:
## Description

Two bugs in `SerializerTask.model_dump_json()`:

### 1. Mutates self permanently

When the serialized size exceeds `max`, the method replaces context attributes on `self`:

```python
self.initial_context = self.size_message
self.current_context = self.size_message
self.previous_context = self.size_message
```

If `model_dump_json()` is called a second time, or if the serializer instance is reused (e.g., first for Telegram notification, then for API persistence), the original context data is **permanently lost**.

### 2. Truncates JSON mid-string (not shown in current code but was in earlier versions)

If truncation produces output that cuts a JSON string mid-value, the result is invalid JSON. Consumers (Telegram, Discord, API) receive broken data.

## Location

`dotflow/core/serializers/task.py`, method `model_dump_json()` (line ~49)

## How to reproduce

```python
from dotflow.core.serializers.task import SerializerTask

task = SerializerTask(
    task_id=0,
    status="Completed",
    initial_context={"large_data": "x" * 10000},
    current_context={"result": "ok"},
    max=500,
)

# First call — truncates and mutates self
json1 = task.model_dump_json()
print(task.initial_context)
# Expected: {"large_data": "xxx..."}
# Actual: "Context size exceeded"

# Second call — original data is gone
json2 = task.model_dump_json()
# Both calls return the truncated version — no way to get original data back
```

## Expected behavior

Use local variables instead of mutating `self`:

```python
def model_dump_json(self, **kwargs) -> str:
    data = json.loads(super().model_dump_json(serialize_as_any=True, **kwargs))
    data["error"] = data["errors"][-1] if data["errors"] else None
    dump_json = json.dumps(data)

    if self.max and len(dump_json) > self.max:
        data["initial_context"] = self.size_message
        data["current_context"] = self.size_message
        data["previous_context"] = self.size_message
        dump_json = json.dumps(data)

    return dump_json
```

This way `self` is never modified and repeated calls always produce correct output.

## Impact

- Telegram/Discord notifications receive corrupted context data after the first truncation
- API persistence stores "Context size exceeded" instead of actual data
- Any code that calls `model_dump_json()` more than once loses data silently

## Priority

High

PR #224, opened 2026-04-09
🪲 BUG-#217: Fix SerializerTask.model_dump_json() mutating self and truncating JSON

FernandoCelmer:
## Description

Fix two bugs in `SerializerTask.model_dump_json()` in `dotflow/core/serializers/task.py`:

1. **Self mutation** — context fields were replaced on `self` instead of the local dict, permanently losing original data on repeated calls
2. **Truncated JSON** — `dump_json[0:self.max]` cut JSON mid-string, producing invalid output

## Motivation and Context

Fixes issue #217 — `model_dump_json()` was mutating `self` and producing invalid truncated JSON

## Changes

- Replace `self.initial_context = ...` with `data["initial_context"] = ...` (operates on local dict)
- Remove `dump_json[0:self.max]` truncation (contexts replaced with size message is enough to reduce size)

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)

## Checklist

- [ ] Tests added
- [ ] CHANGELOG updated
- [ ] Documentation updated

### DIFF
--- a/dotflow/core/serializers/task.py
+++ b/dotflow/core/serializers/task.py
@@ -46,17 +46,10 @@ class SerializerTask(BaseModel):
         dump_json = json.dumps(data)
 
         if self.max and len(dump_json) > self.max:
-            self.initial_context = self.size_message
-            self.current_context = self.size_message
-            self.previous_context = self.size_message
-
-            data = json.loads(
-                super().model_dump_json(serialize_as_any=True, **kwargs)
-            )
-            data["error"] = data["errors"][-1] if data["errors"] else None
+            data["initial_context"] = self.size_message
+            data["current_context"] = self.size_message
+            data["previous_context"] = self.size_message
             dump_json = json.dumps(data)
-
-            return dump_json[0 : self.max]
 
         return dump_json
 
--- a/dotflow/core/serializers/task.py
+++ b/dotflow/core/serializers/task.py
@@ -51,6 +51,11 @@ class SerializerTask(BaseModel):
             data["previous_context"] = self.size_message
             dump_json = json.dumps(data)
 
+            if len(dump_json) > self.max:
+                data["errors"] = []
+                data["error"] = None
+                dump_json = json.dumps(data)
+
         return dump_json
 
     @field_validator("errors", mode="before")
--- a/tests/core/test_serializer_task.py
+++ b/tests/core/test_serializer_task.py
@@ -1,0 +1,91 @@
+"""Test SerializerTask model_dump_json"""
+
+import json
+import unittest
+
+from dotflow.core.context import Context
+from dotflow.core.serializers.task import SerializerTask
+
+
+class TestSerializerTaskDumpJson(unittest.TestCase):
+
+    def _make_task(self, **kwargs):
+        defaults = {
+            "task_id": 0,
+            "_status": "Completed",
+            "_duration": 1.5,
+            "_initial_context": None,
+            "_current_context": None,
+            "_previous_context": None,
+            "group_name": "default",
+            "retry_count": 0,
+            "_errors": [],
+        }
+        defaults.update(kwargs)
+        return SerializerTask(**defaults)
+
+    def test_returns_valid_json(self):
+        task = self._make_task()
+        result = task.model_dump_json()
+
+        parsed = json.loads(result)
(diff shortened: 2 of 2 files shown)


dotflow/core/serializers/task.py
```diff
@@ -46,18 +46,11 @@ def model_dump_json(self, **kwargs) -> str:
...
-            self.previous_context = self.size_message
-
-            data = json.loads(
-                super().model_dump_json(serialize_as_any=True, **kwargs)
-            )
-            data["error"] = data["errors"][-1] if data["errors"] else None
+            data["initial_context"] = self.size_message
+            data["current_context"] = self.size_message
+            data["previous_context"] = self.size_message
             dump_json = json.dumps(data)
```
FernandoCelmer, line 53: [Blocking]

**Problem** — The old `return dump_json[0:self.max]` was correctly removed because it produced invalid JSON by slicing mid-string. However, it was the **only guarantee** that the output never exceeded `max` characters. The new code replaces the three context fields with `size_message` but never re-checks whether the resulting JSON still fits within `max`.

If non-context fields (`errors`, `tracebacks`, or other large payloads) are large enough, the serialized output will silently exceed the `max` limit.

**Failure scenario** — Discord notifications use `max=1024` and Telegram uses `max=4000`. A task with multiple retry errors, each carrying a long traceback, could easily produce a payload that exceeds these limits even after context replacement. The webhook API call would fail with a 400 error, silently dropping the notification.

**Fix** — After replacing contexts, re-check the length and apply secondary reduction if still over `max`:

```python
if self.max and len(dump_json) > self.max:
    data["initial_context"] = self.size_message
    data["current_context"] = self.size_message
    data["previous_context"] = self.size_message
    dump_json = json.dumps(data)

    # If still over max after context replacement, truncate other large fields
    if len(dump_json) > self.max:
        data["errors"] = []
        data["error"] = None
        dump_json = json.dumps(data)

return dump_json
```

This preserves the valid-JSON guarantee (no mid-string slice) while still enforcing `max` as a hard upper bound.

FernandoCelmer (commented):
🔍 Code Review

**Code issues found: 2**

| # | Severity | Comment |
|---|----------|---------|
| 1 | [Blocking] | [Max constraint no longer enforced](https://github.com/dotflow-io/dotflow/pull/224#discussion_r3054845632) |
| 2 | [Suggestion] | [No tests for max parameter path](https://github.com/dotflow-io/dotflow/pull/224#discussion_r3054845635) |

dotflow/core/serializers/task.py
```diff
@@ -46,18 +46,11 @@ def model_dump_json(self, **kwargs) -> str:
...
 
-            return dump_json[0 : self.max]
-
         return dump_json
```
FernandoCelmer, line 60: [Suggestion]

**Problem** — The PR checklist states "I have added tests that prove my fix is effective", but no test files were changed in this PR. The `max` parameter path in `model_dump_json()` has **zero regression test coverage**. Neither the self-mutation fix nor the size-capping behavior is verified by any test.

This is particularly important because the `max` path is actively used by two notification integrations (Discord with `max=1024` and Telegram with `max=4000`), and this PR modifies the core logic of that path.

**Failure scenario** — A future refactor of this method could silently re-introduce the self-mutation bug or break the context replacement logic, with no test to catch it. Since the `max` path is only exercised through the notification integrations, the regression would only surface in production when a task notification fails.

**Fix** — Add tests that cover:

1. **Self-mutation protection**: call `model_dump_json()` twice with `max` — verify `self` attributes are unchanged after both calls
2. **Valid JSON output**: with `max` smaller than the full JSON, verify the return value is valid JSON (parseable by `json.loads`)
3. **Context replacement**: verify that `initial_context`, `current_context`, `previous_context` are replaced with `size_message` when output exceeds `max`
4. **Max enforcement**: verify the output length does not exceed `max` even with large `errors` payloads

```python
def test_model_dump_json_with_max_does_not_mutate_self():
    task = SerializerTask(max=50, ...)
    original_context = task.initial_context
    task.model_dump_json()
    assert task.initial_context == original_context

def test_model_dump_json_with_max_returns_valid_json():
    task = SerializerTask(max=50, ...)
    result = task.model_dump_json()
    json.loads(result)  # should not raise
```

Issue #237, opened 2026-04-09
Feature: Add Alibaba Cloud Function Compute deployer


FernandoCelmer:
## Summary

Add `AliyunFCDeployer` to deploy dotflow pipelines to Alibaba Cloud Function Compute (FC), following the same pattern as AWS Lambda and GCP Cloud Run deployers.

### Deployer
- [ ] Create `dotflow/cloud/alibaba/` module
- [ ] Create `dotflow/cloud/alibaba/deployers/fc.py` with `AliyunFCDeployer`
- [ ] Create `dotflow/cloud/alibaba/deployers/fc_scheduled.py` with `AliyunFCScheduledDeployer`
- [ ] Create `dotflow/cloud/alibaba/constants.py` with platforms, region, credentials
- [ ] Create `dotflow/cloud/alibaba/services/` for CR (Container Registry) and FC operations

### CLI
- [ ] Add `alibaba-fc` and `alibaba-fc-scheduled` to deploy command
- [ ] Add platforms to constants
- [ ] Add `--alibaba-namespace` CLI arg if needed

### Dependencies
- [ ] Add `deploy-alibaba` optional dependency in `pyproject.toml`
- [ ] SDK: `alibabacloud-fc-open20210406`, `alibabacloud-tea-openapi`

### Templates
- [ ] Cloud templates already exist in dotflow-io/templates repo
- [ ] Validate `s.yaml` (Serverless Devs) format

### Documentation
- [ ] Add Alibaba FC to deploy guide
- [ ] Add to release notes

## References

- SDK: `alibabacloud-fc-open20210406`
- Templates: `cloud/alibaba-fc/`, `cloud/alibaba-fc-scheduled/`

PR #238, opened 2026-04-09
⚙️ FEATURE-#237: Add Alibaba Cloud Function Compute deployer

FernandoCelmer:
# Description

Add `AliyunFCDeployer` and `AliyunFCScheduledDeployer` to deploy dotflow pipelines to Alibaba Cloud Function Compute.

Issue: [📌 ISSUE-#237](https://github.com/dotflow-io/dotflow/issues/237)

## Changes

- Add `dotflow/cloud/alibaba/` module (deployers, services, constants)
- Add `alibaba-fc` and `alibaba-fc-scheduled` to deploy command
- Add `deploy-alibaba` optional dependency
- Add cloud docs pages for Alibaba FC
- Update README platforms table with all icons (local SVGs)
- Update integrations page with Cloud Deploy cards
- Standardize all schedule references to cron format
- Add 11 unit tests for Alibaba deployers and CR
- Download all platform icons to docs/assets

## Type of change

- [x] New feature (non-breaking change which adds functionality)
- [x] Documentation updated

# Checklist:

- [x] I have added tests that prove my feature works
- [x] New and existing unit tests pass locally with my changes
- [x] I have made corresponding changes to the documentation


dotflow/cloud/alibaba/services/cr.py
```diff
@@ -0,0 +1,98 @@
...
+            [
+                "docker",
+                "login",
+                "--username",
+                key_id,
+                "--password-stdin",
+                self._registry,
+            ],
+            input=password,
+            check=True,
```
FernandoCelmer, line 75: [Blocking]

**Problem** — `subprocess.run()` is called with `input=password` where `password` is a `str`, but without `text=True`. By default, Python's `subprocess` module operates in binary mode — when `text=True` is omitted, `stdin` expects `bytes`. Passing a `str` raises `TypeError: a bytes-like object is required, not 'str'`. This means `login()` — and therefore every `push()` call — crashes unconditionally.

**Failure scenario** — Running `dotflow deploy --platform alibaba-fc --project my-pipeline` always fails at the Docker login step:

```
TypeError: memoryview: a bytes-like object is required, not 'str'
```

No image is ever pushed, regardless of credentials or Docker state.

**Fix** — Add `text=True` to the `subprocess.run` call, exactly as the AWS ECR counterpart (`dotflow/cloud/aws/services/ecr.py`) does:

```python
run(
    ["docker", "login", "--username", key_id, "--password-stdin", self._registry],
    input=password,
    text=True,
    check=True,
)
```

**References** — [subprocess.run — text mode](https://docs.python.org/3/library/subprocess.html#subprocess.run)

dotflow/cloud/alibaba/deployers/fc.py
```diff
@@ -0,0 +1,132 @@
...
+                    custom_container_config=(
+                        models.CustomContainerConfig(
+                            image=image_uri,
+                            port=9000,
+                        )
+                    ),
+                ),
+            )
+        except Exception:
+            print(f"  {settings.STEP_ICON} Creating function '{name}'...")
```
FernandoCelmer, line 117: [Blocking]

**Problem** — A bare `except Exception` is used to detect whether the function already exists. If `get_function()` fails for any reason other than "function not found" (e.g. network timeout, throttling, authentication error, malformed response), the code silently falls through to `create_function()`. This swallows the real error and attempts an incorrect operation.

**Failure scenario** — An operator deploys with an expired access key. `get_function()` raises a `403 Forbidden`. The code ignores this, calls `create_function()` instead, which also fails with a second `403`. The operator sees "create failed" instead of "invalid credentials" — making the issue very hard to debug, especially in CI pipelines.

```python
# Current behavior:
get_function() → 403 Forbidden (swallowed)
create_function() → 403 Forbidden (confusing error shown)

# Expected behavior:
get_function() → 403 Forbidden → raised immediately with clear message
```

**Fix** — Catch only the specific "not found" exception. The Alibaba FC SDK raises exceptions with status codes. Narrow the except clause:

```python
try:
    self._fc.get_function(name)
    # update path...
except Exception as err:
    if hasattr(err, "status_code") and err.status_code == 404:
        # create path...
    else:
        raise
```

For reference, the AWS deployer catches `self._lambda.exceptions.ResourceNotFoundException` — a specific exception type — so only a genuine "not found" triggers the create path.

dotflow/cloud/alibaba/deployers/fc_scheduled.py
```diff
@@ -0,0 +1,76 @@
...
+                        models.TimerTriggerConfig(
+                            cron_expression=schedule,
+                            payload="{}",
+                            enable=True,
+                        )
+                    ),
+                ),
+            )
+        except Exception:
+            self._fc.create_trigger(
```
FernandoCelmer, line 63: [Blocking]

**Problem** — Same pattern as `_create_or_update_function` in `fc.py`: a bare `except Exception` is used to detect whether the trigger exists. Any non-404 error from `get_trigger()` or `update_trigger()` is silently swallowed, falling through to `create_trigger()`.

**Failure scenario** — The FC service throttles the `get_trigger()` call (rate limit). The code catches the throttling error, then calls `create_trigger()` — which either:
1. Fails with "TriggerAlreadyExists", producing a confusing error, or
2. Succeeds and creates a **duplicate timer trigger**, causing the function to fire twice as often as expected.

```python
# Current behavior:
get_trigger() → 429 Too Many Requests (swallowed)
create_trigger() → creates duplicate trigger or fails with misleading error

# Expected behavior:
get_trigger() → 429 Too Many Requests → raised immediately, deploy retried
```

**Fix** — Apply the same narrowed exception handling:

```python
try:
    self._fc.get_trigger(name, trigger_name)
    self._fc.update_trigger(...)
except Exception as err:
    if hasattr(err, "status_code") and err.status_code == 404:
        self._fc.create_trigger(...)
    else:
        raise
```

FernandoCelmer (commented):
🔍 Code Review

**Code issues found: 3**

| # | Severity | Comment |
|---|----------|---------|
| 1 | [Blocking] | [`subprocess.run` missing `text=True` crashes every deploy](https://github.com/dotflow-io/dotflow/pull/238#discussion_r3060719740) |
| 2 | [Blocking] | [Bare `except Exception` swallows real API errors in FC deployer](https://github.com/dotflow-io/dotflow/pull/238#discussion_r3060719748) |
| 3 | [Blocking] | [Same bare `except Exception` in trigger setup](https://github.com/dotflow-io/dotflow/pull/238#discussion_r3060719753) |

PR #249, opened 2026-04-11
🪲 BUG-#247: Fix _serialize_context crash with non-Context list items

### LINKED ISSUE #247: Bug: _serialize_context crashes when storage is a list of non-Context objects
## Description

When an `@action` returns a plain Python list (e.g. `[dict, dict, ...]`), the `SerializerTask._serialize_context` method crashes with:

```
AttributeError: 'str' object has no attribute 'storage'
```

## Steps to Reproduce

```python
@action
async def fetch_all():
    return [{"id": 1}, {"id": 2}]
```

When `task.result()` is called, the serializer tries to iterate the list and calls `._format_storage(item)` which expects a `Context` object with `.storage`, but receives a raw `dict` (or after `json.dumps`, a `str`).

## Root Cause

In `dotflow/core/serializers/task.py`, `_serialize_context` checks `isinstance(ctx.storage, list)` and iterates items assuming they are `Context` instances:

```python
@classmethod
def _serialize_context(cls, ctx: Context) -> Any:
    if isinstance(ctx.storage, list):
        contexts = {}
        for item in ctx.storage:
            if isinstance(item, Context):
                contexts[item.task_id] = cls._serialize_context(item)
            else:
                contexts[item.task_id] = cls._format_storage(item)  # item has no .task_id or .storage
        return contexts
    return cls._format_storage(ctx)
```

The `else` branch calls `cls._format_storage(item)` which expects a `Context` with `.storage`, but `item` is a raw object (dict, str, etc).

## Expected Behavior

Returning a list from an `@action` should serialize correctly without requiring the list items to be `Context` objects.

## Workaround

Wrap the list in a dict:

```python
@action
async def fetch_all():
    return {"items": [{"id": 1}, {"id": 2}]}
```

## Environment

- dotflow 0.15.0.dev5
- Python 3.14

FernandoCelmer:
## Summary

- Fix `_serialize_context` crash when `ctx.storage` is a list of raw objects (dicts, strings, etc.) instead of `Context` objects
- Fix `context_validator` using truthy check instead of explicit `None` check
- Add support for `tuple` storage serialization
- Add `_format_raw` method for serializing non-Context values
- Configure pytest `testpaths` to exclude `examples/`

## Test plan

- [x] List of dicts serializes without error
- [x] List of strings serializes without error
- [x] List of Context objects still works
- [x] Mixed list (Context + raw) serializes correctly
- [x] Context without task_id in list uses index as key
- [x] Tuple storage serializes correctly
- [x] Empty list returns empty dict
- [x] None context returns None
- [x] Context with None storage returns None
- [x] All 449 existing tests pass

Closes #247


dotflow/core/serializers/task.py
```diff
@@ -93,20 +95,29 @@ def errors_validator(cls, value: list) -> list:
...
 
     @classmethod
     def _serialize_context(cls, ctx: Context) -> Any:
         """Serialize a Context to its storage value."""
-        if isinstance(ctx.storage, list):
+        if isinstance(ctx.storage, (list, tuple)):
             contexts = {}
-            for item in ctx.storage:
+            for index, item in enumerate(ctx.storage):
                 if isinstance(item, Context):
```
FernandoCelmer, line 110: [Blocking]

**Problem** — In `_serialize_context`, Context items are keyed by `task_id` while raw (non-Context) items are keyed by their iteration `index`. When a Context's `task_id` happens to equal the index of an earlier raw item, the Context entry silently **overwrites** the raw entry in the dict — causing permanent data loss.

**Failure scenario** — Consider a mixed list where a raw item sits at index 0 and a Context has `task_id=0`:

```python
from dotflow.core.context import Context

storage = ["raw_value", Context(task_id=0, storage={"nested": True})]

# Iteration:
#   index=0, item="raw_value" (not Context) → contexts[0] = '"raw_value"'
#   index=1, item=Context(task_id=0, ...)    → contexts[0] = {"nested": True}
#
# Result: {0: {"nested": True}}
# "raw_value" is silently lost — overwritten by the Context with task_id=0
```

Another example with multiple collisions:

```python
storage = [
    {"users": 150},                          # index=0 → key=0
    {"orders": 42},                           # index=1 → key=1
    Context(task_id=0, storage={"status": "done"}),  # task_id=0 → key=0 (overwrites!)
    Context(task_id=1, storage={"status": "fail"}),  # task_id=1 → key=1 (overwrites!)
]
# Both raw dicts are silently lost
```

**Fix** — Use a namespace prefix for raw items to prevent key collisions:

```python
# Option A: prefix raw keys
else:
    contexts[f"raw:{index}"] = cls._format_raw(item)

# Option B: always use positional index for both types
for index, item in enumerate(ctx.storage):
    if isinstance(item, Context):
        contexts[index] = cls._serialize_context(item)
    else:
        contexts[index] = cls._format_raw(item)
```

Option B is simpler and guarantees no collisions, but changes the key semantics for Context items (index instead of task_id). Option A preserves task_id keys for Context items while isolating raw keys.

**Note** — The current test `test_list_mixed_context_and_raw` does not catch this because the raw item is at index 1 and the Context has `task_id=5`, so there is no collision. Adding a test like the examples above would reproduce the bug.

dotflow/core/serializers/task.py
```diff
@@ -93,20 +93,25 @@ def errors_validator(cls, value: list) -> list:
...
-                    contexts[item.task_id] = cls._serialize_context(item)
+                    key = item.task_id if item.task_id is not None else index
+                    contexts[key] = cls._serialize_context(item)
                 else:
-                    contexts[item.task_id] = cls._format_storage(item)
+                    contexts[f"raw:{index}"] = cls._format_raw(item)
             return contexts
```
FernandoCelmer, line 118: [Blocking]

**Problem** — When `task_id is None`, the fallback key is `index` (an integer). This integer can collide with the `task_id` of another Context in the same list, silently overwriting data.

**Failure scenario 1** — Two Contexts, one with `task_id=None`:

```python
from dotflow.core.context import Context

storage = [
    Context(task_id=None, storage={"step": "extract"}),  # task_id is None → key = 0 (index fallback)
    Context(task_id=0, storage={"step": "transform"}),    # task_id = 0 → key = 0
]

# Result: {0: {"step": "transform"}}
# The first Context ("extract") is silently lost — overwritten by task_id=0
```

**Failure scenario 2** — Multiple Contexts with `task_id=None` mixed with explicit IDs:

```python
storage = [
    Context(task_id=None, storage={"a": 1}),  # key = 0 (index)
    Context(task_id=None, storage={"b": 2}),  # key = 1 (index)
    Context(task_id=1, storage={"c": 3}),      # key = 1 (task_id) → overwrites "b"!
]

# Result: {0: {"a": 1}, 1: {"c": 3}}
# Context with storage {"b": 2} is silently lost
```

**Failure scenario 3** — Default Context() has `task_id=0`, so the `None` fallback is never actually exercised by `test_list_context_without_task_id` since `Context(storage={"data": True})` already has `task_id=0` as default.

**Fix** — Use a namespace prefix when `task_id` is `None`, consistent with the `raw:{index}` pattern:

```python
if isinstance(item, Context):
    if item.task_id is not None:
        key = item.task_id
    else:
        key = f"ctx:{index}"
    contexts[key] = cls._serialize_context(item)
```

This guarantees no collision between:
- Context keys with explicit task_id → `int` (e.g. `0`, `1`, `5`)
- Context keys without task_id → `str` (e.g. `"ctx:0"`, `"ctx:2"`)
- Raw item keys → `str` (e.g. `"raw:0"`, `"raw:1"`)

PR #253, opened 2026-04-12
🪲 BUG-#245: Fix TimeoutError not caught on Python 3.9/3.10

### LINKED ISSUE #245: concurrent.futures.TimeoutError not caught on Python 3.9/3.10 — timeout silently uncaught
## Description
`_execute_with_timeout` in `core/engine.py` calls `future.result(timeout=seconds)`, which raises `concurrent.futures.TimeoutError` on expiry. The catch block at lines 134 and 176 catches bare `TimeoutError` (the builtin).

On **Python 3.9 and 3.10**, `concurrent.futures.TimeoutError` is a *separate class* that inherits directly from `Exception`, **not** from `builtins.TimeoutError`. This means the timeout is never caught on those versions.

## Location
`dotflow/core/engine.py` — lines 134, 170-178

## Current Behavior
```python
except TimeoutError:
    future.cancel()
    raise TimeoutError()
```
On Python 3.9/3.10, `concurrent.futures.TimeoutError` propagates uncaught, bypasses retry logic, and the task is marked FAILED with a confusing generic error instead of a proper `TimeoutError`.

## Expected Behavior
Both `builtins.TimeoutError` and `concurrent.futures.TimeoutError` should be caught:

```python
from concurrent.futures import TimeoutError as FuturesTimeoutError

except (TimeoutError, FuturesTimeoutError):
    future.cancel()
    raise TimeoutError()
```

## Impact
- **Severity**: Critical
- **Affected versions**: Python 3.9, 3.10
- **Symptom**: Timeout never triggers properly, task crashes with unhandled exception

## Priority
Critical

FernandoCelmer:
## Summary

- Import `concurrent.futures.TimeoutError as FuturesTimeoutError`
- Catch both `TimeoutError` and `FuturesTimeoutError` in `execute_with_retry` (line 136) and `_execute_with_timeout` (line 181)
- Re-raise as `TimeoutError() from None` for clean traceback

## Problem

On Python 3.9/3.10, `concurrent.futures.TimeoutError` is a separate class that does **not** inherit from `builtins.TimeoutError`. The code only caught `TimeoutError` (builtin), so `future.result(timeout=N)` raised an uncaught exception that bypassed retry logic.

On Python 3.11+, they unified the classes so the bug was invisible.

## Tests

- `test_timeout_expired_marks_task_failed` — real timeout with `sleep(5)` and `timeout=1`
- `test_futures_timeout_error_is_caught` — mocks `_execute_with_timeout` to raise `FuturesTimeoutError` directly

## Test plan

- [x] All 3 timeout tests pass
- [x] ruff lint + format pass with `.code_quality/ruff.toml`

Closes #245

### DIFF
--- a/tests/core/test_engine.py
+++ b/tests/core/test_engine.py
@@ -446,11 +446,11 @@ class TestTaskEngine(unittest.TestCase):
 
         self.assertEqual(task.status, TypeStatus.FAILED)
 
-    def test_futures_timeout_error_is_caught(self):
+    def test_futures_timeout_error_converted_to_builtin(self):
         from concurrent.futures import (
             TimeoutError as FuturesTimeoutError,
         )
-        from unittest.mock import patch
+        from unittest.mock import MagicMock, patch
 
         task = Task(
             task_id=0,
@@ -463,14 +463,11 @@ class TestTaskEngine(unittest.TestCase):
             previous_context=Context(),
         )
 
-        with (
-            patch.object(
-                engine,
-                "_execute_with_timeout",
-                side_effect=FuturesTimeoutError("timed out"),
-            ),
-            engine.start(),
-        ):
-            engine.execute_with_retry()
+        mock_future = MagicMock()
+        mock_future.result.side_effect = FuturesTimeoutError("timed out")
 
-        self.assertEqual(task.status, TypeStatus.FAILED)
+        with patch("dotflow.core.engine.ThreadPoolExecutor") as mock_pool:
+            mock_pool.return_value.submit.return_value = mock_future
+
+            with self.assertRaises(TimeoutError):
+                engine._execute_with_timeout(seconds=1)
--- a/tests/core/test_engine.py
+++ b/tests/core/test_engine.py
@@ -419,3 +419,58 @@ class TestTaskEngine(unittest.TestCase):
             engine.execute_with_retry()
 
         self.assertEqual(task.status, TypeStatus.COMPLETED)
+
+    def test_timeout_expired_marks_task_failed(self):
+        from time import sleep as _sleep
+
+        from dotflow import action
+
+        @action(timeout=1)
+        def slow_step():
+            _sleep(5)
+            return {"done": True}
+
+        task = Task(
+            task_id=0,
+            step=slow_step,
+            callback=simple_callback,
+        )
+        engine = TaskEngine(
+            task=task,
+            workflow_id=self.workflow_id,
+            previous_context=Context(),
+        )
+
+        with engine.start():
+            engine.execute_with_retry()
+
+        self.assertEqual(task.status, TypeStatus.FAILED)
+
+    def test_futures_timeout_error_is_caught(self):
+        from concurrent.futures import (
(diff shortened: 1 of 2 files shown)


tests/core/test_engine.py
```diff
@@ -419,3 +419,58 @@ def test_engine_execute_with_timeout(self):
...
+            previous_context=Context(),
+        )
+
+        with engine.start():
+            engine.execute_with_retry()
+
+        self.assertEqual(task.status, TypeStatus.FAILED)
+
+    def test_futures_timeout_error_is_caught(self):
+        from concurrent.futures import (
```
FernandoCelmer, line 450: [Suggestion]

**Problem** — This test patches `_execute_with_timeout` to raise `FuturesTimeoutError` directly. However, the actual fix in `engine.py` (line 181) catches `FuturesTimeoutError` **inside** `_execute_with_timeout` and converts it to a builtin `TimeoutError`. So in production, `FuturesTimeoutError` never escapes `_execute_with_timeout` — it is always converted before reaching `execute_with_retry`.

This means the test exercises a code path that cannot happen in production:

```python
# What happens in production (line 181):
def _execute_with_timeout(self, seconds):
    try:
        return future.result(timeout=seconds)
    except (TimeoutError, FuturesTimeoutError):  # ← caught HERE
        future.cancel()
        raise TimeoutError() from None  # ← converted to builtin

# What the test simulates:
with patch.object(engine, "_execute_with_timeout",
                  side_effect=FuturesTimeoutError("timed out")):
    engine.execute_with_retry()
    # FuturesTimeoutError escapes _execute_with_timeout directly
    # This path is unreachable — the real method converts it first
```

**Impact** — The test gives false confidence about the Python 3.9/3.10 fix. The real fix (line 181) is only covered by `test_timeout_expired_marks_task_failed`, which relies on wall-clock time.

**Fix** — Test `_execute_with_timeout` directly by mocking `future.result()` to raise `FuturesTimeoutError`, then assert the output is a builtin `TimeoutError`:

```python
def test_futures_timeout_error_converted_to_builtin(self):
    from concurrent.futures import TimeoutError as FuturesTimeoutError
    from unittest.mock import MagicMock, patch

    task = Task(task_id=0, step=action_step_with_timeout, callback=simple_callback)
    engine = TaskEngine(
        task=task, workflow_id=self.workflow_id, previous_context=Context()
    )

    mock_future = MagicMock()
    mock_future.result.side_effect = FuturesTimeoutError("timed out")

    with patch("dotflow.core.engine.ThreadPoolExecutor") as mock_pool:
        mock_pool.return_value.__enter__ = MagicMock(return_value=mock_pool)
        mock_pool.return_value.__exit__ = MagicMock(return_value=False)
        mock_pool.return_value.submit.return_value = mock_future

        with self.assertRaises(TimeoutError):
            engine._execute_with_timeout(seconds=1)

    # Confirms FuturesTimeoutError is converted to builtin TimeoutError
```

This directly tests the conversion logic at line 181, which is the actual Python 3.9/3.10 fix.

PR #254, opened 2026-04-12
⚙️ FEATURE-#252: Improve model_dump_json truncation strategy

### LINKED ISSUE #252: Enhancement: Improve SerializerTask.model_dump_json truncation strategy
## Description

The current `model_dump_json(max=N)` truncation in `SerializerTask` works but has several limitations that should be addressed in a future release.

## Current Behavior

```python
def model_dump_json(self, **kwargs) -> str:
    data = self.model_dump(...)
    dump_json = json.dumps(data)

    if not self.max or len(dump_json) <= self.max:
        return dump_json

    # Step 1: replace ALL 3 contexts
    data["initial_context"] = "Context size exceeded"
    data["current_context"] = "Context size exceeded"
    data["previous_context"] = "Context size exceeded"

    # Step 2: if still over, clear errors
    if len(dump_json) > self.max:
        data["errors"] = []
        data["error"] = None
```

### 1. All-or-nothing context truncation
All 3 contexts are replaced even if only 1 is large. A smarter approach would truncate only the largest context first, then the next, etc.

### 2. Polluted data with string message
The API receives `"Context size exceeded"` as a string in JSON columns. This pollutes the database with a sentinel string instead of `null` or a structured indicator.

**Suggestion:** Use `None` or a structured object like `{"truncated": true, "original_size": 15000000}`.

### 3. No final size check
After clearing errors (step 2), the size is not validated again. If some other field is unexpectedly large, the JSON is returned exceeding the limit.

### 4. `max=0` is falsy
`not self.max` on line 61 evaluates `max=0` as `False`, skipping truncation. Should use `self.max is None` instead.

### 5. No partial context truncation
When a context is 15MB but the limit is 5MB, the entire context is replaced. A better approach could keep the first N bytes of the context.

## Proposal

```python
def model_dump_json(self, **kwargs) -> str:
    data = self.model_dump(...)
    dump_json = json.dumps(data)

    if self.max is None or len(dump_json) <= self.max:
        return dump_json

    context_fields = ["current_context", "previous_context", "initial_context"]
    for field in sorted(context_fields, key=lambda f: len(str(data.get(f, ""))), reverse=True):
        data[field] = None
        dump_json = json.dumps(data)
        if len(dump_json) <= self.max:
            return dump_json

    data["errors"] = []
    data["error"] = None
    return json.dumps(data)
```

This truncates contexts one at a time (largest first), uses `None` instead of a string sentinel, and stops as soon as the limit is met.

## Location

`dotflow/core/serializers/task.py` — `model_dump_json` method

FernandoCelmer:
## Summary

- Gradual context truncation — removes largest context first, stops when under limit
- Fix `max=0` falsy bug — `not self.max` → `self.max is None`
- JSON structured message — `{"message": "Context size exceeded"}` instead of plain string
- Remove `size_message` string sentinel, use dict field

## Before vs After

**Before:** All 3 contexts replaced at once, even if only 1 is large
```
15MB payload → replace ALL contexts → clear errors → return
```

**After:** Contexts removed one at a time (largest first)
```
15MB payload → remove current_context (14MB) → 1MB → fits, return
                 ↑ initial and previous preserved
```

## Changes

| Issue | Fix |
|-------|-----|
| All-or-nothing truncation | Gradual: largest context first |
| `max=0` skipped | `self.max is None` check |
| String sentinel `"Context size exceeded"` | JSON `{"message": "Context size exceeded"}` |

## Test plan

- [x] `test_with_max_truncates_largest_context_first` — only largest removed
- [x] `test_with_max_truncates_all_contexts_if_needed` — all removed when necessary
- [x] `test_with_max_zero_truncates` — `max=0` works
- [x] `test_with_max_clears_errors_if_still_over` — errors cleared as last resort
- [x] All 19 serializer tests pass
- [x] ruff lint + format pass

Closes #252

### DIFF
--- a/dotflow/core/serializers/task.py
+++ b/dotflow/core/serializers/task.py
@@ -80,7 +80,14 @@ class SerializerTask(BaseModel):
 
         data["errors"] = []
         data["error"] = None
-        return json.dumps(data)
+        dump_json = json.dumps(data)
+
+        if self.max and len(dump_json) > self.max:
+            for field in context_fields:
+                data[field] = None
+            dump_json = json.dumps(data)
+
+        return dump_json
 
     @field_validator("errors", mode="before")
     @classmethod
--- a/dotflow/core/serializers/task.py
+++ b/dotflow/core/serializers/task.py
@@ -67,11 +67,11 @@ class SerializerTask(BaseModel):
             "previous_context",
             "initial_context",
         ]
-        for field in sorted(
-            context_fields,
-            key=lambda f: len(str(data.get(f, ""))),
-            reverse=True,
-        ):
+
+        def field_size(field: str) -> int:
+            return len(str(data.get(field, "")))
+
+        for field in sorted(context_fields, key=field_size, reverse=True):
             data[field] = self.size_message
             dump_json = json.dumps(data)
             if len(dump_json) <= self.max:
--- a/dotflow/core/serializers/task.py
+++ b/dotflow/core/serializers/task.py
@@ -68,10 +68,11 @@ class SerializerTask(BaseModel):
             "initial_context",
         ]
 
-        def field_size(field: str) -> int:
-            return len(str(data.get(field, "")))
-
-        for field in sorted(context_fields, key=field_size, reverse=True):
+        for field in sorted(
+            context_fields,
+            key=lambda f: len(str(data.get(f, ""))),
+            reverse=True,
+        ):
             data[field] = self.size_message
             dump_json = json.dumps(data)
             if len(dump_json) <= self.max:
--- a/dotflow/core/serializers/task.py
+++ b/dotflow/core/serializers/task.py
@@ -44,8 +44,9 @@ class SerializerTask(BaseModel):
         default_factory=list, alias="_errors"
     )
     max: Optional[int] = Field(default=None, exclude=True)
-    size_message: Optional[str] = Field(
-        default="Context size exceeded", exclude=True
+    size_message: dict = Field(
+        default={"message": "Context size exceeded"},
+        exclude=True,
     )
 
     @computed_field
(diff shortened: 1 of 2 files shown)


dotflow/core/serializers/task.py
```diff
@@ -58,20 +59,28 @@ def model_dump_json(self, **kwargs) -> str:
...
+            reverse=True,
+        ):
+            data[field] = self.size_message
             dump_json = json.dumps(data)
+            if len(dump_json) <= self.max:
+                return dump_json
 
-        return dump_json
+        data["errors"] = []
+        data["error"] = None
```
FernandoCelmer, line 82: [Blocking]

**Problem** — After the gradual truncation loop exhausts all three context fields and falls through to the error-clearing fallback, `json.dumps(data)` is returned **without checking** whether the result now fits within `self.max`. If the base payload (model metadata + errors) already exceeds `max`, the returned string silently violates the size contract.

**Failure scenario 1** — Discord notification with `max=1024`:

```python
from dotflow.core.serializers.task import SerializerTask

# Task with 10 large error tracebacks, no context
task = SerializerTask(
    task_id=1,
    workflow_id="abc",
    status="Failed",
    _errors=[{"attempt": i, "exception": "RuntimeError", "message": "x" * 200, "traceback": "y" * 500} for i in range(10)],
    max=1024,
)

result = task.model_dump_json()

# Gradual loop: all 3 contexts are already None → nothing to truncate
# Fallback: clears errors and error
# But the base payload (with timestamps, status, etc.) may still be > 1024
print(len(result))  # Could exceed 1024 — Discord API rejects it
```

**Failure scenario 2** — `max=0` (newly valid thanks to the `is None` fix):

```python
task = SerializerTask(task_id=1, workflow_id="abc", status="Completed", max=0)
result = task.model_dump_json()

# max=0 means "zero bytes allowed"
# But the base JSON skeleton is ~200+ bytes
# len(result) > 0 → contract violated
```

**Failure scenario 3** — The test `test_with_max_clears_errors_if_still_over` does not assert `len(result) <= task.max`, so this gap is invisible in CI.

**Fix** — Add a final size check after clearing errors. If it still exceeds `max`, either truncate the JSON string or document the contract explicitly:

```python
# Option A: strict guarantee with final truncation
data["errors"] = []
data["error"] = None
dump_json = json.dumps(data)

if len(dump_json) <= self.max:
    return dump_json

# Last resort: truncate to max (caller gets valid-ish JSON prefix)
return dump_json[:self.max]

# Option B: document the contract (best-effort, not strict)
data["errors"] = []
data["error"] = None
return json.dumps(data)  # NOTE: may exceed max if base payload > max
```

Also, add a size assertion to the test:

```python
def test_with_max_clears_errors_if_still_over(self):
    # ... existing setup ...
    result = task.model_dump_json()
    parsed = json.loads(result)
    self.assertEqual(parsed["errors"], [])
    self.assertIsNone(parsed["error"])
    self.assertLessEqual(len(result), task.max)  # ← add this
```

Issue #241, opened 2026-04-11
ServerDefault: update_task sends full payload and create_task crashes on None initial_context


FernandoCelmer:
## Description

Two bugs in `ServerDefault` provider:

1. **`create_task` crashes when `initial_context` is `None`** — accesses `task.initial_context.storage` without null check. The `Context()` getter returns a `Context` object but `.storage` may not be serializable.

2. **`update_task` sends full `task.result()` payload** — includes fields that `TaskUpdate` on the API does not accept (`task_id`, `initial_context`, etc), causing 422 errors.

## Fix applied

- `create_task`: null-safe access on `initial_context`
- `update_task`: send only valid fields (`status`, `errors`, `retry_count`, `duration`, `current_context`, `previous_context`, `started_at`, `finished_at`)
- Added debug logging to `_request`

## Related

- PR #240 (Server provider implementation)

Issue #239, opened 2026-04-09
⚙️ FEATURE: Implement Server provider for remote API communication


FernandoCelmer:
## Summary

Implement a new `Server` ABC provider to replace the current `Api` provider, enabling full bidirectional communication between the dotflow library and a remote API server (e.g., `dotflow-api`).

## Motivation

The current `Api` ABC and its `ApiDefault` implementation are limited:
- `create_workflow` works (best-effort HTTP POST)
- `update_workflow` is a no-op
- `create_task` is a no-op
- `update_task` is a no-op

The dotflow library needs a proper provider to send **real-time execution data** (workflow status, task status, errors, duration, context) to a remote server like `dotflow-api`.

### ABC: `Server`

```python
class Server(ABC):

    @abstractmethod
    def create_workflow(self, workflow: Any) -> None:
        """Register a new workflow on the remote server."""

    @abstractmethod
    def update_workflow(self, workflow: Any) -> None:
        """Update workflow status (in_progress, completed, failed)."""

    @abstractmethod
    def create_task(self, task: Any) -> None:
        """Register a new task under a workflow."""

    @abstractmethod
    def update_task(self, task: Any) -> None:
        """Update task status, duration, context, errors, retry_count."""
```

### Provider: `ServerDefault`

Default HTTP implementation that communicates with `dotflow-api`:

- **Auth** via `X-User-Token` header
- **Config** via env vars: `DOTFLOW_SERVER_URL`, `DOTFLOW_USER_TOKEN`
- **Endpoints used**:
  - `POST /workflows` — create workflow
  - `PATCH /workflows/{id}` — update workflow status
  - `POST /workflows/{id}/tasks` — create task
  - `PATCH /tasks/{id}` — update task (status, duration, errors, context)

### Integration with Config

```python
from dotflow import DotFlow, Config
from dotflow.providers import ServerDefault

config = Config(
    server=ServerDefault(
        base_url="https://api.example.com/api/v1",
        user_token="your-api-token",
    )
)

workflow = DotFlow(config=config)
```

Or via environment variables:

```bash
export DOTFLOW_SERVER_URL=https://api.example.com/api/v1
export DOTFLOW_USER_TOKEN=your-api-token
```

### Task data sent on update

```json
{
  "status": "Completed",
  "duration": 1.23,
  "retry_count": 0,
  "errors": [],
  "initial_context": {},
  "current_context": {"result": "data"},
  "previous_context": {}
}
```

## Tasks

[...]

PR #258, opened 2026-04-12
⚙️ FEATURE-#257: Managed-mode ServerAPI provider and task numbering

### LINKED ISSUE #257: ⚙️ FEATURE: Managed mode — CLI reads WORKFLOW_ID and wires ServerAPI from environment
## Context

The library ships an abstract `Server` provider so that workflow and task status can be reported to a remote HTTP backend in real time. A working reference implementation lives in `examples/server_flow/server_flow/server.py`.

This is enough for a user who writes their own Python entry-point and wires the provider manually. It is **not** enough for a *managed runner* — any orchestrator that launches a container and calls:

```bash
dotflow start --step user_pkg.actions.step_one
```

…expecting status to flow back over HTTP without the user's code having to import anything extra.

A managed runner typically injects these environment variables into the container:

```
SERVER_BASE_URL=https://example.com/api/v1
SERVER_USER_TOKEN=<short-lived token>
WORKFLOW_ID=<UUID already known to the backend>
TASK_ID=<pre-created task row id>
```

For the loop to close, two things have to happen inside that container:

1. **The `DotFlow` instance must use the same `workflow_id` the backend knows about** — otherwise status updates get written to a ghost workflow that the backend never sees.
2. **A `ServerAPI` provider must call back** over HTTP so the remote observer sees `Not started → In progress → Completed` in real time.

Today neither happens automatically. The user would have to hand-roll both for every workflow — which defeats the "connect a repo, click run" UX a managed runner is selling.

---

## Current behaviour (end-to-end trace)

Real container run captured in a managed environment:

- Backend creates a workflow with id `W = 0545d4ca-99d1-4e65-a2eb-90f199773faf`.
- Container is dispatched with the env vars above, including `WORKFLOW_ID=W`.
- `dotflow start --step user_pkg.actions.step_one` runs.
- Logs:
  ```
  INFO [dotflow]: ID 4e5016e5-3f8a-4041-a7b1-c3d9ec6f73b3 - 1 - In progress
  INFO [dotflow]: ID 4e5016e5-3f8a-4041-a7b1-c3d9ec6f73b3 - 1 - Completed
  ```
  The id `4e5016e5-...` is **not** the `W` the backend tracks — it was generated randomly by `DotFlow()`.
- No `PATCH /workflows/{W}/tasks/{tid}` is ever issued because the default `Server` provider is a no-op.
- From the observer's point of view the task stays in its initial state forever while the container has long since exited.

### 1. `dotflow/cli/commands/start.py`

```python
class StartCommand(Command):
    def setup(self):
        workflow = self._new_workflow()
        workflow.task.add(
            step=self.params.step,
            callback=self.params.callback,
            initial_context=self.params.initial_context,
        )
        workflow.start(mode=self.params.mode)

    def _new_workflow(self):
        if not self.params.storage:
            return DotFlow()   # ← no workflow_id, no server config
        ...
        config = Config(storage=storage)
        return DotFlow(config=config)   # ← same problem
```

The CLI does not read `WORKFLOW_ID`, `SERVER_BASE_URL` or `SERVER_USER_TOKEN`. It also never attaches a `Server` provider that talks to a real HTTP endpoint.

### 2. `dotflow/core/dotflow.py`

```python
class DotFlow:
    def __init__(self, config=None, workflow_id=None):
        self.workflow_id = workflow_id or uuid4()      # ← random if not passed
        self._config = config if config else Config()
        self._config.server.create_workflow(workflow=self.workflow_id)   # ← always fires
```

When `workflow_id` is not provided, a random UUID is generated. And `create_workflow` is called unconditionally — even if the caller passed an already-existing id.

### 3. `dotflow/providers/server_default.py`

```python
class ServerDefault(Server):
    def create_workflow(self, workflow): pass
    def update_workflow(self, workflow, status=""): pass
    def create_task(self, task): pass
    def update_task(self, task): pass
```

`Config()` defaults to `ServerDefault`, which is a no-op. So unless the user explicitly wires a real provider, no HTTP calls happen.

### 4. The working reference lives in `examples/`

A working HTTP implementation already exists at `examples/server_flow/server_flow/server.py`:

```python
class ServerAPI(Server):
    def __init__(self, base_url, user_token, timeout=15.0):
        self._base_url = base_url.rstrip("/")
        self._user_token = user_token
        self._timeout = timeout

    def create_workflow(self, workflow):
        self._post(f"{self._base_url}/workflows", json={"id": str(workflow)})

    def update_workflow(self, workflow, status=""):
        self._patch(f"{self._base_url}/workflows/{workflow}", json={"status": status})

    def create_task(self, task):
        data = task.result(max=5_000_000)
        data["id"] = data.pop("task_id", task.task_id)
        self._post(f"{self._base_url}/workflows/{task.workflow_id}/tasks", json=data)

    def update_task(self, task):
        data = task.result(max=5_000_000)
        self._patch(
            f"{self._base_url}/workflows/{task.workflow_id}/tasks/{task.task_id}",
            json=data,
        )
```

The example is excellent for self-hosted scenarios where the user copies the file into their project. It is not usable by a managed runner that cannot expect the user to write any integration code.

---

### 1. Move `ServerAPI` into the library

Create `dotflow/providers/server_api.py` and expose it from `dotflow.providers`:

```python
# dotflow/providers/__init__.py

from dotflow.providers.server_default import ServerDefault
from dotflow.providers.server_api import ServerAPI
from dotflow.providers.storage_default import StorageDefault
from dotflow.providers.storage_file import StorageFile
from dotflow.providers.storage_gcs import StorageGCS
from dotflow.providers.storage_s3 import StorageS3

__all__ = [
    "ServerAPI",
    "ServerDefault",
    "StorageDefault",
    "StorageFile",
    "StorageGCS",
    "StorageS3",
]
```

```python
# dotflow/providers/server_api.py

"""ServerAPI — sends workflow/task events to a remote HTTP endpoint."""

from __future__ import annotations

import logging
from typing import Any

from requests import patch, post
from requests.exceptions import RequestException

from dotflow.abc.server import Server

logger = logging.getLogger(__name__)


class ServerAPI(Server):
    """HTTP implementation of the Server ABC.

    Posts workflow creation and task status updates to a remote API
    using a user token in the ``X-User-Token`` header. Failures are
    logged and swallowed so a temporary network glitch never crashes
    a running workflow.
    """

    def __init__(
        self,
        base_url: str,
        user_token: str,
        timeout: float = 15.0,
    ):
        self._base_url = base_url.rstrip("/")
        self._user_token = user_token
        self._timeout = timeout

    @property
    def _headers(self) -> dict:
        return {
            "X-User-Token": self._user_token,
            "Content-Type": "application/json",
        }

    def _post(self, url: str, json: dict) -> None:
        try:
            response = post(
                url,
                json=json,
                headers=self._headers,
                timeout=self._timeout,
            )
            logger.info(
                "POST %s [%s]",
                url,
                response.status_code,
            )
        except RequestException as error:
            logger.error("POST %s failed: %s", url, error)

    def _patch(self, url: str, json: dict) -> None:
        try:
            response = patch(
                url,
                json=json,
                headers=self._headers,
                timeout=self._timeout,
            )
            logger.info(
                "PATCH %s [%s]",
                url,
                response.status_code,
            )
        except RequestException as error:
            logger.error("PATCH %s failed: %s", url, error)

    def create_workflow(self, workflow: Any) -> None:
        self._post(
            f"{self._base_url}/workflows",
            json={"id": str(workflow)},
        )

    def update_workflow(self, workflow: Any, status: str = "") -> None:
        self._patch(
            f"{self._base_url}/workflows/{workflow}",
            json={"status": status},
        )

    def create_task(self, task: Any) -> None:
        data = task.result(max=5_000_000)
        data["id"] = data.pop("task_id", task.task_id)
        self._post(
            f"{self._base_url}/workflows/{task.workflow_id}/tasks",
            json=data,
        )

    def update_task(self, task: Any) -> None:
        data = task.result(max=5_000_000)
        self._patch(
            f"{self._base_url}/workflows/{task.workflow_id}/tasks/{task.task_id}",
            json=data,
        )
```

### 2. Make `DotFlow.__init__` skip `create_workflow` when the id comes from outside

When a managed runner provides the workflow id, the workflow row already exists on the remote side. Calling `create_workflow` from inside the container would either duplicate the row or fail with a unique-id violation. The safest contract:

> If the caller passed an explicit `workflow_id`, assume the workflow already exists on the server and do **not** call `create_workflow`.

```python
# dotflow/core/dotflow.py

class DotFlow:
    def __init__(
        self,
        config: Config | None = None,
        workflow_id: str | None = None,
    ) -> None:
        self._externally_provided_id = workflow_id is not None
        self.workflow_id = workflow_id or uuid4()
        self._config = config if config else Config()

        if not self._externally_provided_id:
            self._config.server.create_workflow(workflow=self.workflow_id)

        self.task = TaskBuilder(
            config=self._config, workflow_id=self.workflow_id
        )

        self.start = partial(
            Manager,
            tasks=self.task.queue,
            workflow_id=self.workflow_id,
            config=self._config,
        )

        self.schedule = partial(
            self._config.scheduler.start, workflow=self.start
        )
```

Existing usage without an explicit `workflow_id` keeps working identically — the behaviour change is strictly additive.

### 3. CLI auto-wires managed mode from environment

`StartCommand` reads three optional env vars. If present, it activates managed mode:

```python
# dotflow/cli/commands/start.py

import os
from os import system

from dotflow import Config, DotFlow
from dotflow.cli.command import Command
from dotflow.core.types.execution import TypeExecution
from dotflow.providers import (
    ServerAPI,
    ServerDefault,
    StorageDefault,
    StorageFile,
    StorageGCS,
    StorageS3,
)


class StartCommand(Command):
    def setup(self):
        workflow = self._new_workflow()

        workflow.task.add(
            step=self.params.step,
            callback=self.params.callback,
            initial_context=self.params.initial_context,
        )

        workflow.start(mode=self.params.mode)

        if self.params.mode == TypeExecution.BACKGROUND:
            system("/bin/bash")

    def _new_workflow(self):
        config = self._build_config()
        workflow_id = os.getenv("WORKFLOW_ID")
        return DotFlow(config=config, workflow_id=workflow_id)

    def _build_config(self):
        storage = self._build_storage()
        server = self._build_server()

        if storage is None and server is None:
            return None
        return Config(
            storage=storage or StorageDefault(),
            server=server or ServerDefault(),
        )

    def _build_server(self):
        base_url = os.getenv("SERVER_BASE_URL")
        user_token = os.getenv("SERVER_USER_TOKEN")
        if not base_url or not user_token:
            return None
        return ServerAPI(base_url=base_url, user_token=user_token)

    def _build_storage(self):
        if not self.params.storage:
            return None

        storage_classes = {
            "default": StorageDefault,
            "file": StorageFile,
            "s3": StorageS3,
            "gcs": StorageGCS,
        }
        storage_cls = storage_classes.get(self.params.storage)
        storage_with_path = {StorageDefault, StorageFile}

        if storage_cls in storage_with_path:
            return storage_cls(path=self.params.path)
        return storage_cls()
```

Key properties:
- `dotflow start --step ...` with no env vars keeps getting a no-op `ServerDefault` (unchanged behaviour).
- With `SERVER_BASE_URL` + `SERVER_USER_TOKEN` set, a real `ServerAPI` is wired automatically.
- With `WORKFLOW_ID` set, that id is reused everywhere — logs, internal state, and HTTP callbacks all reference the same UUID the remote observer is watching.

---

## Behaviour matrix

| Env var set | Stand-alone usage | Managed runner usage |
|---|---|---|
| `WORKFLOW_ID` | no | yes |
| `SERVER_BASE_URL` | no | yes |
| `SERVER_USER_TOKEN` | no | yes |
| `create_workflow` fires on start | yes | no |
| `update_task` fires after each task | no (`ServerDefault`) | yes (`ServerAPI`) |
| Workflow id in logs | random UUID | matches `WORKFLOW_ID` |

---

## Acceptance criteria

- [ ] `dotflow.providers.ServerAPI` is importable and documented
- [ ] `dotflow start --step X` with no env vars behaves exactly as today
- [ ] `dotflow start --step X` with `WORKFLOW_ID` set uses that id as `workflow.workflow_id`
- [ ] `dotflow start --step X` with `SERVER_BASE_URL` and `SERVER_USER_TOKEN` set emits one HTTP call per task finish: `PATCH {base}/workflows/{wid}/tasks/{tid}`
- [ ] `dotflow start --step X` with `WORKFLOW_ID` set **does not** call `POST /workflows` (no ghost workflow is created)
- [ ] Network failures on the server provider are logged and swallowed — they never abort the running workflow
- [ ] Unit tests cover each of the four matrix rows above
- [ ] `examples/server_flow/` continues to work unchanged (explicit-wire pattern remains supported)

---

### Unit

- `ServerAPI` constructor builds the expected headers and URL.
- `ServerAPI.update_task` sends the right URL shape with `workflow_id` and `task_id` from the task object.
- `DotFlow(workflow_id="abc")` does not call `server.create_workflow`.
- `DotFlow()` (no id) calls `server.create_workflow` exactly once.
- `StartCommand._build_server` returns `None` when either env var is missing, returns a `ServerAPI` when both are set.
- `StartCommand._new_workflow` passes `os.environ["WORKFLOW_ID"]` into `DotFlow` when set.

### Integration

- Spin up a local HTTP stub that accepts `PATCH /workflows/{wid}/tasks/{tid}` and records calls.
- Set `SERVER_BASE_URL`, `SERVER_USER_TOKEN`, `WORKFLOW_ID` env vars.
- Run `dotflow start --step tests.fixtures.dummy:workflow`.
- Assert the stub got at least one PATCH whose URL contains the `WORKFLOW_ID` and whose body contains the expected terminal status.

---

## Migration and backwards compatibility

- Self-hosted usage without any of the three env vars sees identical behaviour.
- Anyone already passing `workflow_id=` explicitly in Python code keeps working — the only delta is that `create_workflow` is no longer called automatically in that case. The workflow is assumed to exist on the server, which matches the common case (checkpoint resume, managed runner). If a legacy script depended on the old implicit creation, the upgrade note should instruct them to either drop the explicit `workflow_id` or to call `server.create_workflow` themselves.
- `examples/server_flow/` stays in the repo as a reference for the explicit-wire pattern.

---

## Non-goals

Explicitly out of scope for this issue (tracked separately):

- Streaming live stdout/stderr from the container to a remote observer (would need WebSocket/SSE; today only final task status is reported).
- Automatic retry with backoff on transient `ServerAPI` failures — current behaviour is log + continue, which is intentional for a first pass.
- Per-task `TASK_ID` injection from env. The library already generates task ids; the remote backend should honour whatever the lib reports rather than pre-creating rows.

---

## References

- Existing reference provider: [`examples/server_flow/server_flow/server.py`](https://github.com/dotflow-io/dotflow/blob/main/examples/server_flow/server_flow/server.py).

FernandoCelmer:
## Description
Closes #257

Enables managed execution: the CLI auto-wires a `ServerAPI` provider from env vars and DotFlow continues task numbering correctly when an external `workflow_id` is supplied.

### Modified files/modules:
- `dotflow/constants.py` (new) — Constants like `INITIAL_TASK_ID = 1`
- `dotflow/abc/server.py` (modified) — Abstract base class for Server providers
- `dotflow/providers/server_api.py` (new) — ServerAPI HTTP provider implementing authentication and task ID management
- `dotflow/core/dotflow.py` (modified) — Updated `DotFlow.__init__` to handle external `workflow_id`
- `dotflow/core/task.py` (modified) — `TaskBuilder` accepts `initial_task_id` parameter
- `dotflow/cli/commands/start.py` (modified) — Auto-configures ServerAPI provider from env vars
- `dotflow/cli/commands/cloud.py`, `dotflow/cli/commands/deploy.py` (modified) — CLI updates
- `dotflow/__init__.py`, `dotflow/providers/__init__.py` (modified) — Module exports
- `tests/providers/test_server_api.py` (new) — Tests for ServerAPI provider
- `tests/cli/test_start_command.py`, `tests/core/test_dotflow.py` (modified) — Updated tests
- `pyproject.toml`, `poetry.lock`, `LAST_VERSION` (modified) — Dependencies and version

## Motivation and Context
Enables managed workflow execution where the CLI auto-wires a `ServerAPI` provider from environment variables. This allows DotFlow to continue task numbering correctly when an external `workflow_id` is supplied, supporting integration with dotflow-api for centralized workflow management.

## Types of changes
- [ ] Documentation

## Checklist

## Test plan
- [x] `poetry run ruff check dotflow/ tests/`
- [x] `poetry run pytest -x -q` — 488 passed, 4 skipped
- [ ] Manually run the CLI with `WORKFLOW_ID`, `SERVER_BASE_URL` and `SERVER_USER_TOKEN` set against a staging dotflow-api and confirm task ids continue from the last run


dotflow/providers/server_api.py
```diff
@@ -0,0 +1,105 @@
...
+            json={"status": status},
+        )
+
+    def create_task(self, task: Any) -> None:
+        data = task.result(max=5_000_000)
+        data["id"] = data.pop("task_id", task.task_id)
+        self._post(
+            f"{self._base_url}/workflows/{task.workflow_id}/tasks",
+            json=data,
+        )
```
FernandoCelmer: [Blocking]

**Problem** — `update_task` sends the **raw, unfiltered** `task.result()` dictionary to the PATCH endpoint. This payload includes fields like `task_id`, `workflow_id`, `group_name`, `created_at`, and `error` — all of which are **not expected** by the API for an update operation.

This is the **exact same pattern** that caused **BUG-#241**: `ServerDefault.update_task` previously sent the full `task.result()` output, and the API returned **422 Unprocessable Entity** errors because of unexpected fields. The fix in `ServerDefault` was to allowlist only the fields the API accepts.

**Failure scenario** — Consider a task that completes successfully:

```python
# task.result(max=5_000_000) returns something like:
{
    "task_id": 42,
    "workflow_id": "abc-123",
    "group_name": "default",
    "created_at": "2026-04-12T10:00:00",
    "status": "Completed",
    "duration": 1.5,
    "errors": [],
    "retry_count": 0,
    "current_context": {...},
    "previous_context": {...},
    "started_at": "2026-04-12T10:00:00",
    "finished_at": "2026-04-12T10:00:01",
    "error": null
}
```

This entire dict is sent as the PATCH body. If the API validates incoming fields strictly (as it did for `ServerDefault`), every `update_task` call will silently fail with a 422 — and because `_patch` only catches `RequestException` (network errors), the 422 HTTP response won't even be logged as an error (see Issue #4 below).

**Fix** — Apply the same field allowlist that was established in `ServerDefault` after BUG-#241:

```python
def update_task(self, task: Any) -> None:
    result = task.result(max=5_000_000)
    data = {
        k: result.get(k)
        for k in (
            "status",
            "errors",
            "retry_count",
            "duration",
            "current_context",
            "previous_context",
            "started_at",
            "finished_at",
        )
        if result.get(k) is not None
    }
    self._patch(
        (
            f"{self._base_url}/workflows/{task.workflow_id}"
            f"/tasks/{task.task_id}"
        ),
        json=data,
    )
```

This ensures only the fields the API expects are included in the payload, matching the convention already validated in production via `ServerDefault`.

FernandoCelmer (commented):
🔍 Code Review

**Code issues found: 5**

| # | Severity | Comment |
|---|----------|---------|
| 1 | [Blocking] | [Unfiltered payload in update_task repeats BUG-#241](https://github.com/dotflow-io/dotflow/pull/258#discussion_r3070036177) |
| 2 | [Blocking] | [Silent task_id fallback to 1 risks collisions](https://github.com/dotflow-io/dotflow/pull/258#discussion_r3070036181) |
| 3 | [Blocking] | [Blocking update calls contradict PR #244 pattern](https://github.com/dotflow-io/dotflow/pull/258#discussion_r3070036182) |
| 4 | [Suggestion] | [HTTP 4xx/5xx errors silently swallowed](https://github.com/dotflow-io/dotflow/pull/258#discussion_r3070036183) |
| 5 | [Suggestion] | [Stale docstrings for workflow_id and TaskBuilder](https://github.com/dotflow-io/dotflow/pull/258#discussion_r3070036184) |

dotflow/providers/server_api.py
```diff
@@ -0,0 +1,105 @@
...
+    def get_next_task_id(self, workflow: Any) -> int:
+        url = f"{self._base_url}/workflows/{workflow}/tasks/next-id"
+        try:
+            response = get(
+                url,
+                headers=self._headers,
+                timeout=self._timeout,
+            )
+            logger.info("GET %s [%s]", url, response.status_code)
+            if response.status_code == 200:
```
FernandoCelmer: [Blocking]

**Problem** — `get_next_task_id` checks `response.status_code == 200` which is narrower than necessary, and **silently falls back** to `INITIAL_TASK_ID = 1` on any non-200 response. This fallback is only logged at `INFO` level, making it effectively invisible in production logs.

The core risk is **task ID collision**: when a resumed workflow silently restarts numbering from 1, it creates duplicate task IDs on the server, potentially overwriting existing task records.

**Failure scenario** — Imagine a workflow with 50 completed tasks is being resumed:

```
# Server is temporarily returning 503 (maintenance window)
GET /workflows/abc-123/tasks/next-id → 503 Service Unavailable

# get_next_task_id logs at INFO level (easily missed):
# INFO: GET https://api.example.com/workflows/abc-123/tasks/next-id [503]

# Falls back silently to INITIAL_TASK_ID = 1
# New tasks are created with task_id=1, 2, 3...
# These OVERWRITE the existing tasks 1-50 on the server!
```

Another edge case: if the API returns `201 Created` or `204 No Content` (both valid 2xx), the code treats them as failures and falls back to 1.

**Fix** — Use `response.ok` (covers all 2xx) and escalate the fallback log to `WARNING`:

```python
def get_next_task_id(self, workflow: Any) -> int:
    url = f"{self._base_url}/workflows/{workflow}/tasks/next-id"
    try:
        response = get(
            url,
            headers=self._headers,
            timeout=self._timeout,
        )
        logger.info("GET %s [%s]", url, response.status_code)
        if response.ok:
            return int(response.json().get("next_id", INITIAL_TASK_ID))
        logger.warning(
            "GET %s returned %s — falling back to task_id=%s. "
            "This may cause task ID collisions if the workflow "
            "already has tasks on the server.",
            url, response.status_code, INITIAL_TASK_ID,
        )
    except (RequestException, ValueError, TypeError) as error:
        logger.error("GET %s failed: %s", url, error)

    return INITIAL_TASK_ID
```

This ensures operators are alerted when the fallback activates, giving them a chance to intervene before data is corrupted.

dotflow/providers/server_api.py
```diff
@@ -0,0 +1,105 @@
...
+    def create_workflow(self, workflow: Any) -> None:
+        self._post(
+            f"{self._base_url}/workflows",
+            json={"id": str(workflow)},
+        )
+
+    def update_workflow(self, workflow: Any, status: str = "") -> None:
+        self._patch(
+            f"{self._base_url}/workflows/{workflow}",
+            json={"status": status},
```
FernandoCelmer: [Blocking]

**Problem** — `update_workflow` and `update_task` make **synchronous HTTP calls** (with a 15-second timeout) directly on the caller's thread. This contradicts the **fire-and-forget pattern** established in PR #244, where `ServerDefault` wraps update calls in `Thread(daemon=True).start()` to prevent workflow stalling.

The rationale from PR #244 was clear: `create_*` calls must be synchronous (the record must exist before the next step), but `update_*` calls are status updates that should never block the workflow executor.

**Failure scenario** — Consider a 10-task workflow running against a server with 2-second network latency:

```
# Each task triggers 2 update calls: update_task (start) + update_task (end)
# Plus 2 update_workflow calls (start + end)

# With ServerDefault (PR #244 pattern):
#   Updates fire in daemon threads → workflow completes in ~10s

# With ServerAPI (current implementation):
#   Each _patch() blocks for up to 15s on timeout
#   10 tasks × 2 updates × 2s latency = 40s of pure HTTP wait
#   Worst case (timeouts): 10 × 2 × 15s = 300s added to workflow runtime
```

This is especially problematic because tests mock the HTTP layer, so CI won't catch the performance regression.

**Fix** — Follow the same daemon-thread pattern used in `ServerDefault`:

```python
from threading import Thread

def update_workflow(self, workflow: Any, status: str = "") -> None:
    Thread(
        target=self._patch,
        args=(f"{self._base_url}/workflows/{workflow}",),
        kwargs={"json": {"status": status}},
        daemon=True,
    ).start()

def update_task(self, task: Any) -> None:
    data = task.result(max=5_000_000)  # also apply allowlist per Issue #1
    url = (
        f"{self._base_url}/workflows/{task.workflow_id}"
        f"/tasks/{task.task_id}"
    )
    Thread(
        target=self._patch,
        args=(url,),
        kwargs={"json": data},
        daemon=True,
    ).start()
```

Using `daemon=True` ensures the threads don't prevent process exit, matching the established convention.

dotflow/providers/server_api.py
```diff
@@ -0,0 +1,105 @@
...
+        try:
+            response = post(
+                url,
+                json=json,
+                headers=self._headers,
+                timeout=self._timeout,
+            )
+            logger.info("POST %s [%s]", url, response.status_code)
+        except RequestException as error:
+            logger.error("POST %s failed: %s", url, error)
```
FernandoCelmer: [Suggestion]

**Problem** — `_post` and `_patch` only catch `RequestException` (network-level failures like timeouts, DNS errors, connection refused). **HTTP error responses** (4xx/5xx) are received successfully at the network level, so they bypass the `except` block entirely. The status code is logged at `INFO` level regardless of whether it's 200 or 500, making server-side errors invisible.

This creates a dangerous blind spot: if the Bearer token is rotated, expired, or misconfigured, every API call returns `401 Unauthorized` — but the workflow continues as if everything succeeded.

**Failure scenario** — Operator rotates the API token and restarts the CLI:

```python
# All subsequent calls return 401:
# INFO: POST https://api.example.com/workflows [401]
# INFO: POST https://api.example.com/workflows/abc/tasks [401]  
# INFO: PATCH https://api.example.com/workflows/abc/tasks/1 [401]

# No ERROR or WARNING in logs.
# Workflow "completes successfully" with exit code 0.
# But zero data was persisted to the remote API.
# Operator discovers hours later that all workflow data is missing.
```

**Fix** — Check `response.ok` and log non-2xx responses at `ERROR` level:

```python
def _post(self, url: str, json: dict) -> None:
    try:
        response = post(
            url,
            json=json,
            headers=self._headers,
            timeout=self._timeout,
        )
        if response.ok:
            logger.info("POST %s [%s]", url, response.status_code)
        else:
            logger.error(
                "POST %s [%s]: %s",
                url, response.status_code, response.text[:200],
            )
    except RequestException as error:
        logger.error("POST %s failed: %s", url, error)
```

Apply the same pattern to `_patch`. Including `response.text[:200]` in the error log provides the server's error message without flooding logs with large response bodies.

dotflow/core/dotflow.py
```diff
@@ -5,6 +5,7 @@
 from functools import partial
 from uuid import uuid4
 
+from dotflow.constants import INITIAL_TASK_ID
 from dotflow.core.config import Config
```
FernandoCelmer, line 8: [Suggestion]

**Problem** — This PR significantly changes the semantics of `workflow_id` but **does not update the docstrings** in two places:

**1. `DotFlow.__init__` docstring** (this file, line ~35):
The current docstring says:
```
workflow_id (Optional[str]): Fixed workflow identifier for checkpoint
    resume. If not provided, a random UUID is generated.
```
After this PR, providing `workflow_id` now means: skip `create_workflow`, call `get_next_task_id` to resume task numbering, and operate in **managed mode**. This is a materially different contract than "checkpoint resume". A developer reading the docstring would not know that passing a `workflow_id` suppresses the `create_workflow` call and alters task ID sequencing.

**Suggested update:**
```python
workflow_id (Optional[str]): External workflow identifier for
    managed-mode execution. When provided, ``create_workflow`` is
    skipped (the workflow is assumed to already exist on the remote
    server) and the next task ID is fetched via
    ``server.get_next_task_id()``. If not provided, a random UUID
    is generated and a new workflow is created on the server.
```

**2. `TaskBuilder.__init__` Args** (`dotflow/core/task.py`, line ~304):
The `Args:` section documents `config` and `workflow_id` but omits the new `initial_task_id` parameter, breaking the existing convention of documenting all constructor arguments.

**Suggested addition:**
```python
Args:
    config (Config): Configuration class.
    workflow_id (UUID): Workflow ID.
    initial_task_id (int): Starting task ID for numbering.
        Defaults to ``INITIAL_TASK_ID`` (1). When resuming
        a managed-mode workflow, this is fetched from the
        remote server to avoid ID collisions.
```

Keeping docstrings aligned with behavior prevents confusion for contributors and consumers of the library API.

dotflow/providers/server_api.py
```diff
@@ -0,0 +1,88 @@
...
+        data = task.result(max=5_000_000)
+        data["id"] = data.pop("task_id", task.task_id)
+        self._post(
+            f"{self._base_url}/workflows/{task.workflow_id}/tasks",
+            json=data,
+        )
+
+    def update_task(self, task: Any) -> None:
+        data = task.result(max=5_000_000)
+        self._patch(
```
FernandoCelmer: [Blocking]

**Problem** — `create_task` renames `task_id` to `id` before sending (`data["id"] = data.pop("task_id", task.task_id)`), but `update_task` sends the raw serialized data without this rename. The remote API receives `"id"` on create and `"task_id"` on update for the same resource type.

**Failure scenario** — The remote API creates a task with `{"id": "01ARZ..."}`. A subsequent `update_task` PATCH sends `{"task_id": "01ARZ...", ...}`. If the API schema validates strict field names, the request is rejected. If it silently ignores unknown fields, the task identity is lost in the update payload.

**Fix** — Apply the same rename in `update_task`:

```python
def update_task(self, task: Any) -> None:
    data = task.result(max=5_000_000)
    data["id"] = data.pop("task_id", task.task_id)  # consistent with create_task
    self._patch(
        f"{self._base_url}/workflows/{task.workflow_id}/tasks/{task.task_id}",
        json=data,
    )
```

dotflow/core/engine.py
```diff
@@ -275,7 +277,7 @@ def _execution_with_class(self, class_instance: Callable):
                     previous_context=previous_context,
                     task=self.task,
                 )
-                subcontext.task_id = index
+                subcontext.task_id = str(ULID())
                 new_context.storage.append(subcontext)
```
FernandoCelmer, line 282: [Blocking]

**Problem** — Both the `try` and `except` branches independently call `str(ULID())`, generating different IDs for the same logical task step. Before this PR, both branches used the same `index` variable (introduced in FEATURE-#239), ensuring the task_id was identical regardless of which branch was taken.

**Failure scenario** — When `ExecutionWithClassError` is raised, the subcontext stored in `new_context.storage` gets a different ULID than what would have been assigned in the happy path. Any downstream code (serialization, server reporting, log correlation) that relies on a stable task_id for a given step will see inconsistent values across retry/fallback executions.

**Fix** — Generate the ULID once before the try/except:

```python
task_ulid = str(ULID())
try:
    subcontext = new_object(...)
    subcontext.task_id = task_ulid
    ...
except Exception as error:
    if not isinstance(error, ExecutionWithClassError):
        raise error
    subcontext = new_object(class_instance, ...)
    subcontext.task_id = task_ulid  # same ID
```

FernandoCelmer (commented):
🔍 Code Review

**Code issues found: 2**

| # | Severity | Comment |
|---|----------|---------|
| 1 | [Blocking] | create_task/update_task field naming inconsistency |
| 2 | [Blocking] | Double ULID on class execution error path |

PR #260, opened 2026-04-13
⚙️ FEATURE-#259: Add --workflow flag to dotflow start

### LINKED ISSUE #259: RFC: Run full workflows via CLI with --workflow
## Summary

Today `dotflow start --step module.func` runs a **single** `@action` step. To execute a workflow composed of multiple `task.add(...)` calls you have to wrap everything inside an `@action`-decorated function that calls `.start()` itself — which mixes step and orchestrator semantics and makes the CLI unable to control the execution mode.

This RFC proposes a new, explicit entry point: `--workflow`.

## Motivation

**Current pain** — projects like `dotflow-cloud` end up with:

```python
@action
def main():
    workflow = DotFlow(config=Config())
    workflow.task.add(step=step_one)
    workflow.task.add(step=step_two)
    workflow.start(mode="sequential")
    return workflow
```

Problems:
- `@action` semantically means *one step*; here it hides a whole
  orchestrator
- `.start()` is hardcoded inside; the CLI can't override mode,
  storage, etc.
- `--step` resolves this as a step and would execute it as one task,
  not as a pipeline

## Proposal

Add `--workflow` (or `-w`) to the `start` command. It resolves a factory `module.attr` that returns a `DotFlow` instance, and the CLI is responsible for starting it.

### Usage

```bash
dotflow start --workflow dotflow_cloud.workflow.main
dotflow start --workflow dotflow_cloud.workflow.main --mode parallel
```

### Factory shape

```python
from dotflow import Config, DotFlow
from dotflow_cloud.actions import step_one, step_two

def main() -> DotFlow:
    workflow = DotFlow(config=Config())
    workflow.task.add(step=step_one)
    workflow.task.add(step=step_two)
    return workflow  # no @action, no .start() here
```

### CLI behavior

- `--step` and `--workflow` become **mutually exclusive** (one required)
- With `--workflow`: import factory → call it → assert it returned a `DotFlow` → call `workflow.start(mode=self.params.mode)`
- `--mode`, `--storage`, `--path` continue to work (CLI wins over factory defaults when explicitly passed)

### Proposed changes

1. `dotflow/cli/setup.py` — add `--workflow` in a mutually exclusive group with `--step`
2. `dotflow/cli/commands/start.py` — branch on `params.workflow`, new private `_start_from_factory()` using `Module(self.params.workflow)`
3. `dotflow/core/exception.py` — new `InvalidWorkflowFactory` raised when the resolved attr is not callable or does not return `DotFlow`
4. Docs + example in `docs_src/` showing the factory pattern

## Open questions

1. **Override semantics** — should CLI flags (`--mode`, `--storage`) always override factory config, or only when the factory did not set them? Recommendation: CLI always wins
2. **Dotted attribute support** — `Module` already supports `pkg.module.attr` via `rpartition`, so no change there
3. **Schedule parity** — should `dotflow schedule --workflow ...` get the same treatment?
4. **Return-type validation** — should we allow factories that return `None` and call `.start()` themselves (backwards compat), or strictly require a returned `DotFlow`?

## Non-goals

- No change to `@action` semantics
- No deprecation of `--step` — it remains the way to run a single task

FernandoCelmer:
Closes #259

## Description

Adds a `--workflow` / `-w` flag to `dotflow start` as a mutually exclusive alternative to `--step`, enabling users to run full multi-step pipelines from the CLI without wrapping orchestration logic inside an `@action` decorator.

### Modified files

| File | Change |
|------|--------|
| `dotflow/cli/commands/start.py` | Split `setup()` into `_start_from_step()` and `_start_from_factory()`. Factory path resolves the target via `Module(path)`, validates it is callable and returns a `DotFlow` instance, then calls `result.start(mode=...)`. Flags exclusive to `--step` (`--callback`, `--initial-context`) raise `WorkflowFlagConflict` if passed with `--workflow`. |
| `dotflow/cli/setup.py` | Made `--step` and `--workflow` a `add_mutually_exclusive_group(required=True)`. Registered exception handlers for `InvalidWorkflowFactory` and `WorkflowFlagConflict`. Fixed `set_defaults` to target the top-level parser instead of the argument group. |
| `dotflow/core/exception.py` | Added `InvalidWorkflowFactory` (non-callable or wrong return type) and `WorkflowFlagConflict` (step-only flags used with `--workflow`) with descriptive messages. |
| `tests/cli/test_start_command.py` | New test suite covering: non-callable factory, wrong return type, `--callback` flag conflict, `--initial-context` flag conflict, and happy-path factory execution. |
| `docs/nav/how-to/cli/with-workflow.md` | New how-to guide explaining `--step` vs `--workflow`, factory shape, and CLI examples. |
| `docs_src/cli/cli_with_workflow.py` | Example demonstrating both single-step and factory-based workflow invocation. |
| `docs/nav/examples/cli.md` | Added `cli_with_workflow` to the examples table. |
| `docs/nav/how-to/cli/simple-start.md` | Added note linking to `--workflow` as an alternative to `--step`. |
| `docs/nav/how-to/cli/with-deploy.md` | Modernized note block syntax from `!!!` to `///`. |
| `docs/nav/how-to/index.md` | Updated CLI section description to mention `--workflow`. |
| `docs/nav/reference/exception.md` | Added API reference for `InvalidWorkflowFactory` and `WorkflowFlagConflict`. |
| `mkdocs.yml` | Registered `with-workflow.md` in navigation and redirects. |

## Motivation and Context

Before this PR, running a multi-step workflow from the CLI required wrapping the entire orchestration inside an `@action`-decorated function — mixing step and orchestrator semantics. The CLI had no way to control execution mode, storage, or other parameters for multi-step pipelines.

The `--workflow` flag solves this by accepting a factory function (`module.attr`) that returns a configured `DotFlow` instance. The CLI is then responsible for calling `.start()`, giving it full control over execution parameters like `--mode`.

## Types of changes

- [x] Documentation

## Checklist

- [x] I have updated the documentation accordingly

### DIFF
--- a/dotflow/cli/commands/start.py
+++ b/dotflow/cli/commands/start.py
@@ -4,7 +4,7 @@ from os import system
 
 from dotflow import Config, DotFlow
 from dotflow.cli.command import Command
-from dotflow.core.exception import InvalidWorkflowFactory
+from dotflow.core.exception import InvalidWorkflowFactory, WorkflowFlagConflict
 from dotflow.core.module import Module
 from dotflow.core.types.execution import TypeExecution
 from dotflow.providers import (
@@ -13,6 +13,7 @@ from dotflow.providers import (
     StorageGCS,
     StorageS3,
 )
+from dotflow.utils.basic_functions import basic_callback
 
 
 class StartCommand(Command):
@@ -37,6 +38,14 @@ class StartCommand(Command):
         workflow.start(mode=self.params.mode)
 
     def _start_from_factory(self):
+        step_only_flags = {
+            "--callback": self.params.callback is not basic_callback,
+            "--initial-context": self.params.initial_context is not None,
+        }
+        for flag, provided in step_only_flags.items():
+            if provided:
+                raise WorkflowFlagConflict(flag=flag)
+
         factory = Module(self.params.workflow)
 
         if not callable(factory):
--- a/dotflow/cli/commands/start.py
+++ b/dotflow/cli/commands/start.py
@@ -4,6 +4,8 @@ from os import system
 
 from dotflow import Config, DotFlow
 from dotflow.cli.command import Command
+from dotflow.core.exception import InvalidWorkflowFactory
+from dotflow.core.module import Module
 from dotflow.core.types.execution import TypeExecution
 from dotflow.providers import (
     StorageDefault,
@@ -14,7 +16,17 @@ from dotflow.providers import (
 
 
 class StartCommand(Command):
+
     def setup(self):
+        if getattr(self.params, "workflow", None):
+            self._start_from_factory()
+        else:
+            self._start_from_step()
+
+        if self.params.mode == TypeExecution.BACKGROUND:
+            system("/bin/bash")
+
+    def _start_from_step(self):
         workflow = self._new_workflow()
 
         workflow.task.add(
@@ -25,8 +37,18 @@ class StartCommand(Command):
 
         workflow.start(mode=self.params.mode)
 
-        if self.params.mode == TypeExecution.BACKGROUND:
-            system("/bin/bash")
+    def _start_from_factory(self):
+        factory = Module(self.params.workflow)
+
+        if not callable(factory):
+            raise InvalidWorkflowFactory(factory=self.params.workflow)
+
+        result = factory()
+
+        if not isinstance(result, DotFlow):
+            raise InvalidWorkflowFactory(factory=self.params.workflow)
+
+        result.start(mode=self.params.mode)
 
     def _new_workflow(self):
         if not self.params.storage:
--- a/dotflow/cli/commands/start.py
+++ b/dotflow/cli/commands/start.py
@@ -16,7 +16,6 @@ from dotflow.providers import (
 
 
 class StartCommand(Command):
-
     def setup(self):
         if getattr(self.params, "workflow", None):
             self._start_from_factory()
--- a/docs/nav/examples/cli.md
+++ b/docs/nav/examples/cli.md
@@ -8,3 +8,4 @@
 | [cli_with_mode](https://github.com/dotflow-io/examples/blob/master/cli_with_mode.py) | `dotflow start --step examples.cli_with_mode.simple_step --mode sequential` |
 | [cli_with_output_context](https://github.com/dotflow-io/examples/blob/master/cli_with_output_context.py) | `dotflow start --step examples.cli_with_output_context.simple_step --storage file` |
 | [cli_with_path](https://github.com/dotflow-io/examples/blob/master/cli_with_path.py) | `dotflow start --step examples.cli_with_path.simple_step --path .storage --storage file` |
+| [cli_with_workflow](https://github.com/dotflow-io/examples/blob/master/cli_with_workflow.py) | `dotflow start --workflow examples.cli_with_workflow.pipeline` |
--- a/docs/nav/how-to/cli/simple-start.md
+++ b/docs/nav/how-to/cli/simple-start.md
@@ -7,3 +7,7 @@ dotflow start --step docs_src.basic.simple_cli.simple_step
 ```
 
 {* ./docs_src/basic/simple_cli.py *}
+
+/// note
+`--step` runs a single `@action` function. To run a full multi-step pipeline, use [`--workflow`](with-workflow.md) instead.
+///
--- a/docs/nav/how-to/cli/with-deploy.md
+++ b/docs/nav/how-to/cli/with-deploy.md
@@ -79,11 +79,13 @@ dotflow deploy --platform github-actions --project my-pipeline
 | `--region` | Cloud region (default: us-east-1 for AWS, us-central1 for GCP, cn-hangzhou for Alibaba) |
 | `--schedule` | Cron expression for scheduled platforms (e.g. `*/5 * * * *`) |
 
-!!! note
-    Cron expressions use standard 5-field format (`min hour day month weekday`). Dotflow converts to the cloud provider format automatically (e.g. AWS EventBridge `cron()`).
+/// note
+Cron expressions use standard 5-field format (`min hour day month weekday`). Dotflow converts to the cloud provider format automatically (e.g. AWS EventBridge `cron()`).
+///
 
-!!! note
-    - AWS: `pip install dotflow[deploy-aws]`
-    - GCP: `pip install dotflow[deploy-gcp]`
-    - Alibaba: `pip install dotflow[deploy-alibaba]`
-    - GitHub: `pip install dotflow[deploy-github]`
+/// note
+- AWS: `pip install dotflow[deploy-aws]`
+- GCP: `pip install dotflow[deploy-gcp]`
+- Alibaba: `pip install dotflow[deploy-alibaba]`
+- GitHub: `pip install dotflow[deploy-github]`
+///
--- a/docs/nav/how-to/cli/with-workflow.md
+++ b/docs/nav/how-to/cli/with-workflow.md
@@ -1,0 +1,33 @@
+# With Workflow
+
+The `start` command supports two mutually exclusive entry points:
+
+| Flag | When to use |
+|------|-------------|
+| `--step` / `-s` | Run a single `@action`-decorated function |
+| `--workflow` / `-w` | Run a factory that returns a `DotFlow` instance |
+
+## Single step
+
+```bash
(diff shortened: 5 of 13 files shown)


FernandoCelmer (commented):
🔍 Code Review

**Code issues found: 2**

| # | Severity | Comment |
|---|----------|---------|
| 1 | [Blocking] | CLI flags silently ignored with --workflow |
| 2 | [Suggestion] | No tests for factory validation paths |

dotflow/cli/commands/start.py
```diff
@@ -25,8 +36,18 @@ def setup(self):
...
+
+        if not callable(factory):
+            raise InvalidWorkflowFactory(factory=self.params.workflow)
+
+        result = factory()
+
+        if not isinstance(result, DotFlow):
+            raise InvalidWorkflowFactory(factory=self.params.workflow)
+
+        result.start(mode=self.params.mode)
```
FernandoCelmer, line 59: [Blocking]

**Problem** — `_start_from_factory()` only forwards `self.params.mode` to the resolved workflow. The CLI flags `--callback`, `--initial-context`, `--storage`, and `--path` are still **accepted by argparse** (they belong to the shared argument group), but they are **silently discarded** in this code path. The user receives no error and no warning — the flags simply have no effect.

Compare this with `_start_from_step()`, which wires every flag through:

```python
# _start_from_step (existing) — all flags are connected:
workflow.task.add(
    step=Module(self.params.step),
    callback=Module(self.params.callback),
    initial_context=self.params.initial_context,
)
config = self._build_config()  # uses self.params.storage, self.params.path
workflow.start(mode=self.params.mode)
```

```python
# _start_from_factory (new) — only mode is forwarded:
result = factory()
result.start(mode=self.params.mode)
# ← callback, initial_context, storage, path are all ignored
```

**Failure scenario** — A team has a shared pipeline factory and individual developers customize behavior via CLI flags:

```bash
# Developer A wants to monitor task execution via Slack alerts:
dotflow start --workflow myproject.pipeline --callback myproject.alerts.notify_slack
# → notify_slack is NEVER called. No warning. Silent monitoring gap.

# Developer B wants to persist workflow state to disk:
dotflow start --workflow myproject.pipeline --storage file --path /tmp/dotflow-state
# → StorageDefault (in-memory) is used. State is lost on exit. No warning.

# Developer C passes an initial context for parameterized runs:
dotflow start --workflow myproject.pipeline --initial-context '{"env": "staging"}'
# → Context is ignored. Factory runs with its hardcoded defaults.
```

In all three cases, the CLI **accepts** the flags without error, the workflow **runs successfully**, and the user has **no indication** that their flags were ignored. This is particularly dangerous for `--callback` in production monitoring scenarios — a silent gap in alerting can go unnoticed for days.

**Fix** — Option A (recommended): Emit a warning when incompatible flags are passed with `--workflow`:

```python
def _start_from_factory(self):
    factory = Module(self.params.workflow)

    if not callable(factory):
        raise InvalidWorkflowFactory(factory=self.params.workflow)

    # Warn about flags that are ignored in factory mode
    ignored = []
    if self.params.callback is not basic_callback:
        ignored.append("--callback")
    if self.params.initial_context:
        ignored.append("--initial-context")
    if self.params.storage:
        ignored.append("--storage")
    if self.params.path:
        ignored.append("--path")

    if ignored:
        flags = ", ".join(ignored)
        print(
            f"⚠️  Warning: {flags} ignored when using --workflow. "
            "Configure these inside your factory function instead."
        )

    result = factory()
    if not isinstance(result, DotFlow):
        raise InvalidWorkflowFactory(factory=self.params.workflow)

    result.start(mode=self.params.mode)
```

Option B: Move the incompatible flags into the `--step` exclusive group so argparse rejects them at parse time. This is stricter but requires restructuring the argument groups.

Option C: Apply the CLI flags to the factory result after creation:

```python
result = factory()
if not isinstance(result, DotFlow):
    raise InvalidWorkflowFactory(factory=self.params.workflow)

# Forward CLI overrides to the factory-built workflow
if self.params.callback is not basic_callback:
    for task in result.task._tasks:
        task.callback = Module(self.params.callback)

config = self._build_config()  # builds Config from --storage, --path
result.config = config
result.start(mode=self.params.mode)
```

Option A is the safest — it preserves the factory's autonomy while making the behavior transparent.

dotflow/cli/commands/start.py
```diff
@@ -25,8 +36,18 @@ def setup(self):
 
         workflow.start(mode=self.params.mode)
 
-        if self.params.mode == TypeExecution.BACKGROUND:
-            system("/bin/bash")
+    def _start_from_factory(self):
+        factory = Module(self.params.workflow)
+
+        if not callable(factory):
```
FernandoCelmer, line 51: [Suggestion]

**Problem** — `_start_from_factory()` introduces three distinct validation paths — (1) the `callable()` check, (2) the `isinstance(result, DotFlow)` check, and (3) the happy path — but **zero tests** are added in this PR. This is the core correctness guarantee of the feature: if any of these guards break silently, users will see confusing tracebacks instead of the clear `InvalidWorkflowFactory` message.

Every other exception in this codebase has dedicated tests. For example, `tests/core/test_exception.py` covers `MissingActionDecorator` and `ExecutionModeNotExist` with explicit message assertions. `InvalidWorkflowFactory` follows the same pattern but has no coverage.

Additionally, both failure paths (not callable and wrong return type) raise the **same exception with the same message**. Without tests that exercise each path independently, there is no safety net if a future refactor accidentally removes one of the guards.

**Failure scenario** — Consider a future change where `Module.__new__` is refactored to always return a callable wrapper (e.g., for lazy loading). The `callable()` check would become a no-op — always passing — and users who point `--workflow` at a non-callable attribute would get a confusing `TypeError` from `factory()` instead of the friendly `InvalidWorkflowFactory` message. Without a test that explicitly exercises the non-callable path, this regression ships undetected.

Another scenario: a refactor of `DotFlow` introduces a base class or protocol. The `isinstance` check silently breaks for subclasses or protocol-conforming objects. Without a test for the happy path, there is no regression guard.

**Fix** — Add tests covering all three paths in `tests/cli/` or `tests/core/test_exception.py`:

```python
# tests/cli/test_start_command.py (or similar)

from unittest import TestCase
from unittest.mock import patch, MagicMock
from dotflow import DotFlow
from dotflow.core.exception import InvalidWorkflowFactory

class TestStartFromFactory(TestCase):

    def test_non_callable_factory_raises_invalid_workflow(self):
        """When --workflow points to a non-callable (e.g., a string constant),
        InvalidWorkflowFactory should be raised with the module path."""
        cmd = self._build_command(workflow="myproject.config.VERSION")
        with patch("dotflow.cli.commands.start.Module", return_value="1.0.0"):
            with self.assertRaises(InvalidWorkflowFactory) as ctx:
                cmd._start_from_factory()
            self.assertIn("myproject.config.VERSION", str(ctx.exception))

    def test_factory_returning_wrong_type_raises_invalid_workflow(self):
        """When the factory returns something other than DotFlow,
        InvalidWorkflowFactory should be raised."""
        cmd = self._build_command(workflow="myproject.pipeline.broken")
        with patch("dotflow.cli.commands.start.Module", return_value=lambda: {"not": "a workflow"}):
            with self.assertRaises(InvalidWorkflowFactory) as ctx:
                cmd._start_from_factory()
            self.assertIn("myproject.pipeline.broken", str(ctx.exception))

    def test_factory_returning_dotflow_starts_successfully(self):
        """Happy path: factory returns a DotFlow instance, .start() is called
        with the mode from CLI params."""
        mock_workflow = MagicMock(spec=DotFlow)
        cmd = self._build_command(workflow="myproject.pipeline.main", mode="parallel")
        with patch("dotflow.cli.commands.start.Module", return_value=lambda: mock_workflow):
            cmd._start_from_factory()
            mock_workflow.start.assert_called_once_with(mode="parallel")

    def test_exception_message_matches_pattern(self):
        """InvalidWorkflowFactory message should include the factory path
        for debugging, matching the pattern of other exceptions."""
        from dotflow.core.exception import MESSAGE_INVALID_WORKFLOW_FACTORY
        exc = InvalidWorkflowFactory(factory="my.module.pipeline")
        self.assertEqual(
            exc.args[0],
            MESSAGE_INVALID_WORKFLOW_FACTORY.format(factory="my.module.pipeline"),
        )
```

These four tests provide a regression safety net for the feature's core contract and take less than 5 minutes to write. They also serve as **living documentation** of the expected behavior for future contributors who encounter `--workflow` for the first time.

PR #263, opened 2026-04-14
🪲 BUG-#262: Fix entry point separator to use standard module.path:attribute format

### LINKED ISSUE #262: Fix entry point format to use standard module.path:attribute convention
## Problem

The `Module.import_module()` method in `dotflow/core/module.py` uses `rpartition(".")` to split the entry point string:

```python
module_path, _, attr_name = value.rpartition(".")
```

This forces the format `module.path.attribute` (all dots), which diverges from the Python packaging standard used by every major Python tool (Gunicorn, Uvicorn, Celery, setuptools, pyproject.toml).

## Standard format

The Python ecosystem standard is `module.path:attribute` — dots for the module path, colon as separator before the attribute:

| Tool | Example |
|---|---|
| Gunicorn | `main:app` |
| Uvicorn | `package.module:app` |
| Celery | `proj.celery:app` |
| setuptools | `mymodule:main` |
| Python spec | `importable.module:object.attr` |

Reference: https://packaging.python.org/specifications/entry-points/

## Fix

Change `rpartition(".")` to `rpartition(":")` in `dotflow/core/module.py`:

```python
module_path, _, attr_name = value.rpartition(":")
```

## Impact

- CLI usage changes from `dotflow start --step my_module.my_step` to `dotflow start --step my_module:my_step`
- Breaking change — requires updating all examples and docs

FernandoCelmer:
## Description
Changes the entry point separator in  from  to  to align with Python's standard packaging convention for module entry points. This change affects how the CLI parses and executes module attributes.

**Modified files:**
-  - Updated partition logic
- Test files - Updated to use new  format

## Motivation and Context
This change standardizes dotflow's entry point format to match Python's industry-standard module entry point specification (PEP 440 and setuptools conventions). The colon separator () is the standard delimiter used by Python packaging tools, making dotflow more aligned with ecosystem conventions.

Closes #262

## Types of changes
- [x] Bug fix (change that fixes an issue)
- [ ] New feature (change which adds functionality)
- [x] Breaking change

## Breaking Change Notice
**CLI usage changes:**
- Old format: `dotflow start --step my_module.my_step`
- New format: `dotflow start --step my_module:my_step`

## Checklist
- [ ] I have updated the documentation accordingly (open issue)

### DIFF
--- a/dotflow/core/module.py
+++ b/dotflow/core/module.py
@@ -14,7 +14,7 @@ class Module:
 
     @classmethod
     def import_module(cls, value: str):
-        module_path, _, attr_name = value.rpartition(".")
+        module_path, _, attr_name = value.rpartition(":")
 
         if not module_path:
             raise ImportModuleError(module=value)
--- a/dotflow/core/module.py
+++ b/dotflow/core/module.py
@@ -14,7 +14,8 @@ class Module:
 
     @classmethod
     def import_module(cls, value: str):
-        module_path, _, attr_name = value.rpartition(":")
+        separator = ":" if ":" in value else "."
+        module_path, _, attr_name = value.rpartition(separator)
 
         if not module_path:
             raise ImportModuleError(module=value)
--- a/tests/core/test_task.py
+++ b/tests/core/test_task.py
@@ -148,6 +148,14 @@ class TestTaskSetter(unittest.TestCase):
         with self.assertRaises(ImportModuleError):
             Task(task_id=0, step=input_value, callback=simple_callback)
 
+    def test_set_step_with_dot_format_fallback_success(self):
+        input_value = "tests.mocks.step_function.action_step"
+        expected_value = Action
+
+        task = Task(task_id=0, step=input_value, callback=simple_callback)
+
+        self.assertIsInstance(task.step, expected_value)
+
     def test_set_step_with_function_success(self):
         input_value = action_step
         expected_value = Action
--- a/tests/core/test_task.py
+++ b/tests/core/test_task.py
@@ -135,7 +135,7 @@ class TestTaskSetter(unittest.TestCase):
         self.content = {"foo": "bar"}
 
     def test_set_step_with_path_module_success(self):
-        input_value = "tests.mocks.step_function.action_step"
+        input_value = "tests.mocks.step_function:action_step"
         expected_value = Action
 
         task = Task(task_id=0, step=input_value, callback=simple_callback)
@@ -143,7 +143,7 @@ class TestTaskSetter(unittest.TestCase):
         self.assertIsInstance(task.step, expected_value)
 
     def test_set_step_with_path_module_fail(self):
-        input_value = "tests.mocks.step_function.XPTO"
+        input_value = "tests.mocks.step_function:XPTO"
 
         with self.assertRaises(ImportModuleError):
             Task(task_id=0, step=input_value, callback=simple_callback)
@@ -163,7 +163,7 @@ class TestTaskSetter(unittest.TestCase):
             Task(task_id=0, step=input_value, callback=simple_callback)
 
     def test_set_callback_with_path_module_success(self):
-        input_value = "tests.mocks.step_function.action_step"
+        input_value = "tests.mocks.step_function:action_step"
         expected_value = Action
 
         task = Task(task_id=0, step=action_step, callback=input_value)
@@ -171,13 +171,13 @@ class TestTaskSetter(unittest.TestCase):
         self.assertIsInstance(task.step, expected_value)
 
     def test_set_callback_with_path_module_fail(self):
-        input_value = "tests.mocks.step_function.XPTO"
+        input_value = "tests.mocks.step_function:XPTO"
 
         with self.assertRaises(ImportModuleError):
             Task(task_id=0, step=action_step, callback=input_value)
 
     def test_set_callback_with_path_module_not_callable_fail(self):
-        input_value = "tests.mocks.constants.NOT_CALLABLE"
+        input_value = "tests.mocks.constants:NOT_CALLABLE"
 
         with self.assertRaises(NotCallableObject):
             Task(task_id=0, step=action_step, callback=input_value)
--- a/docs/nav/examples/cli.md
+++ b/docs/nav/examples/cli.md
@@ -2,10 +2,10 @@
 
 | Example | Command |
 | ------- | ------- |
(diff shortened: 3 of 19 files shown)


dotflow/core/module.py
```diff
@@ -14,7 +14,7 @@ def __new__(cls, value: Any):
 
     @classmethod
     def import_module(cls, value: str):
-        module_path, _, attr_name = value.rpartition(".")
+        module_path, _, attr_name = value.rpartition(":")
 
         if not module_path:
             raise ImportModuleError(module=value)
```
FernandoCelmer, line 21: [Blocking]

**Problem** — When a user passes the old dot-only format (e.g., `"my.module.attribute"`), `rpartition(":")` returns `("", "", "my.module.attribute")` — `module_path` is empty and `ImportModuleError` is raised. The error message says the module couldn't be imported, implying the module doesn't exist. In reality, the problem is a **wrong separator**, not a missing module.

**Failure scenario** — A user migrating from the previous version passes:
```python
Task(step="my_package.tasks.process_data")  # old format with dot
```
They receive:
```
ImportModuleError: Error importing Python module 'my_package.tasks.process_data'.
```
They check their installed packages, verify the module path is correct, and waste significant time before realising they just need a `:` instead of the last `.`.

**Fix** — Add an explicit format check before splitting, so the error clearly tells users what went wrong:
```python
@classmethod
def import_module(cls, value: str):
    if ":" not in value:
        raise ImportModuleError(
            module=value,
            message=(
                f"Invalid entry point format: '{value}'. "
                "Expected 'module.path:attribute' (colon as separator). "
                f"Did you mean '{value.rsplit('.', 1)[0]}:{value.rsplit('.', 1)[-1]}'?"
            )
        )
    module_path, _, attr_name = value.rpartition(":")
    ...
```
This surfaces the root cause immediately and can even suggest the correct format to the user.

**References** — Python packaging entry points spec: https://packaging.python.org/en/latest/specifications/entry-points/

tests/core/test_task.py
```diff
@@ -135,15 +135,15 @@ def setUp(self):
...
+        input_value = "tests.mocks.step_function:action_step"
         expected_value = Action
 
         task = Task(task_id=0, step=input_value, callback=simple_callback)
 
         self.assertIsInstance(task.step, expected_value)
 
     def test_set_step_with_path_module_fail(self):
-        input_value = "tests.mocks.step_function.XPTO"
+        input_value = "tests.mocks.step_function:XPTO"
```
FernandoCelmer, line 147: [Suggestion]

**Problem** — This test verifies that an **unknown attribute** raises `ImportModuleError`. However, after this PR, passing the **old dot-only format** (e.g., `"tests.mocks.step_function.XPTO"`) also raises `ImportModuleError` — but for a completely different reason (empty `module_path` from `rpartition(":")`). The test passes regardless of whether the error is "attribute not found" or "separator missing", masking two distinct failure modes behind the same exception.

**Failure scenario** — If a future contributor accidentally reverts the `rpartition` change, `"tests.mocks.step_function:XPTO"` would now split correctly into `module_path="tests.mocks.step_function"` and `attr_name="XPTO"`, and `ImportError` would be raised when `getattr` can't find `XPTO`. The test still passes. But if the contributor had introduced a bug where the separator check is skipped, a dot-only string would also pass through — and the test would still pass, silently hiding the regression.

**Fix** — Add a separate, explicit test for the malformed-format case, and strengthen the existing test to assert the error message content:
```python
def test_set_step_with_path_module_fail(self):
    """Attribute does not exist — correct format, wrong attribute name."""
    input_value = "tests.mocks.step_function:XPTO"
    with self.assertRaises(ImportModuleError):
        Task(task_id=0, step=input_value, callback=simple_callback)

def test_set_step_with_missing_separator_fail(self):
    """Old dot-only format — no colon separator — should fail with a clear message."""
    input_value = "tests.mocks.step_function.action_step"  # dot format, no colon
    with self.assertRaises(ImportModuleError) as ctx:
        Task(task_id=0, step=input_value, callback=simple_callback)
    self.assertIn(":", str(ctx.exception))  # error should mention the expected format
```

FernandoCelmer (commented):
🔍 Code Review

**Code issues found: 3**

| # | Severity | Comment |
|---|----------|---------|
| 1 | [Blocking] | [Docs and examples use stale dot format](https://github.com/dotflow-io/dotflow/pull/263#issuecomment-4240796581) |
| 2 | [Blocking] | [ImportModuleError gives misleading message](https://github.com/dotflow-io/dotflow/pull/263#discussion_r3076710018) |
| 3 | [Suggestion] | [Test does not distinguish separator error from attribute error](https://github.com/dotflow-io/dotflow/pull/263#discussion_r3076710020) |

FernandoCelmer:
[Blocking]

**Problem** — This PR changes the entry point separator from `.` to `:` in `Module.import_module()`, but all documentation and runnable example files still use the old dot-only format. After this PR merges, **every documented example will immediately fail** with `ImportModuleError`.

**Files not updated:**

*Runnable examples (`docs_src/`):*
- `docs_src/basic/simple_cli.py` — calls `os.system("dotflow start --step docs_src.basic.simple_cli.simple_step")`
- `docs_src/cli/cli_with_callback.py` — `os.system("dotflow start --step docs_src.cli.cli_with_callback.simple_step --callback docs_src.cli.cli_with_callback.callback")`
- `docs_src/cli/cli_with_initial_context.py`, `cli_with_mode.py`, `cli_with_output_context.py`, `cli_with_workflow.py`

*Documentation guides (`docs/`):*
- `docs/nav/how-to/cli/simple-start.md`
- `docs/nav/how-to/cli/with-callback.md`
- `docs/nav/how-to/cli/with-initial-context.md`
- `docs/nav/how-to/cli/with-custom-path.md`
- `docs/nav/how-to/cli/with-workflow.md`
- `docs/nav/examples/cli.md`

**Failure scenario** — A new user opens the docs, copies the first example, and runs:
```sh
dotflow start --step docs_src.basic.simple_cli.simple_step
```
They get:
```
ImportModuleError: Error importing Python module 'docs_src.basic.simple_cli.simple_step'.
```
No hint that they need to change it to `docs_src.basic.simple_cli:simple_step`.

**Fix** — Update all affected files to use the colon separator. Each `os.system()` call and every CLI snippet in the docs needs to change from:
```sh
# Before (broken after this PR)
dotflow start --step my.module.my_function

# After (correct)
dotflow start --step my.module:my_function
```

This is a blocking issue because the docs and runnable examples are the primary onboarding path for new users.

PR #271, opened 2026-04-15
⚙️ FEATURE-#269: Add dotflow login with device flow and config file

### LINKED ISSUE #269: Add dotflow login command with device authorization flow + config file loader
## Context

Users today must manually copy API tokens from the dashboard and export `SERVER_USER_TOKEN`; this replaces the copy-paste with a browser-based device flow (same pattern as `gh auth login`, `vercel login`, `fly auth login`).

The API endpoints and the authorize page in the frontend will be handled in separate issues within their respective repositories.

### `dotflow login` command

- `POST {base_url}/auth/cli/device` → receive `{device_code, user_code, verification_uri, interval, expires_in}`.
- Open browser at `verification_uri` via `webbrowser.open`.
- Poll `POST {base_url}/auth/cli/token` with `device_code` every `interval` seconds until:
  - `200` → save token to config, print success.
  - `410` → abort with clear message.
  - `expires_in` elapsed → abort.
- Non-interactive fallback: `dotflow login --token <t>` for CI/CD environments without a browser.

### Config file loader

New location: `~/.dotflow/config.toml`

```toml
[cloud]
base_url = "https://host.dotflow.io/api/v1"
token = "dtf_sk_..."
```

### Resolution precedence

`settings.py` / `providers/server_default.py` must resolve token and base_url in this order:

1. Explicit function args (unchanged).
2. Env vars `SERVER_USER_TOKEN` / `SERVER_BASE_URL` — **win over config file** so CI/CD keeps working.
3. `~/.dotflow/config.toml`.
4. Nothing → server features disabled (current behavior).

### `dotflow logout`

Deletes the token from the config file (does not call the API — token stays valid until rotated). Symmetric to `login`.

## Out of scope

- Workspace concept in config (single token assumed).
- `dotflow init --cloud` — tracked separately.

## Test plan

- Mock device flow end-to-end (start, authorize, token received, config written).
- Precedence tests: env > config.
- `logout` removes token cleanly.
- Non-interactive `--token` path.

## Priority

High — depends on the API endpoint issue; unblocks the wizard on the frontend.

FernandoCelmer:
## Description
Add dotflow login with device flow and config file management

## Summary
- `dotflow login` opens the browser, shows the user_code, polls `/auth/cli/token`, and saves the resulting API token to `~/.dotflow/config.toml`.
- `dotflow logout` removes the saved config.
- New `dotflow/core/config_file.py` with TOML-minimal reader/writer and an env-over-file `resolve()` helper.
- `ServerDefault` now resolves `SERVER_BASE_URL` / `SERVER_USER_TOKEN` via `resolve()` so env wins and the config file is the fallback — CI/CD keeps working when env vars are set, and interactive users get the saved token.
- `dotflow login` tries `https://www.host.dotflow.io/api/v1` first, then `https://www.lib.dotflow.io/api/v1` on `RequestException`. `--base-url` / `SERVER_BASE_URL` skip the fallback.
- Non-interactive `dotflow login --token <t>` bypasses the browser for CI.

## Motivation and Context
Closes #269

## Types of changes

## Test plan
- [x] `ruff check` + `ruff format --check`.
- [x] `pytest tests/cli tests/core` — includes 16 new tests (login device flow, base-url fallback, explicit-wins, logout noop/cleanup, config file roundtrip, 0600 perms, env-over-file precedence).

## Checklist
- [x] I have updated the CHANGELOG
- [x] I have updated the documentation accordingly

Pairs with:
- dotflow-io/dotflow-api#121 — backend endpoints
- dotflow-io/dotflow-frontend#127 — authorize page

### DIFF
--- a/dotflow/cli/commands/login.py
+++ b/dotflow/cli/commands/login.py
@@ -13,9 +13,9 @@ from dotflow.cli.command import Command
 from dotflow.core.config_file import save_cloud_config
 from dotflow.settings import Settings as settings
 
-DEFAULT_BASE_URL = "https://www.cli.dotflow.io/api/v1"
-DEVICE_ENDPOINT = "/auth/cli/device"
-TOKEN_ENDPOINT = "/auth/cli/token"
+DEFAULT_BASE_URL = "https://www.host.dotflow.io/api/v1"
+DEVICE_ENDPOINT = "/cli/device"
+TOKEN_ENDPOINT = "/cli/token"
 TIMEOUT = 15
 DEFAULT_INTERVAL = 5
 
--- a/dotflow/cli/commands/login.py
+++ b/dotflow/cli/commands/login.py
@@ -21,7 +21,6 @@ DEFAULT_INTERVAL = 5
 
 
 class LoginCommand(Command):
-
     def setup(self):
         token = getattr(self.params, "token", None)
         base_url = self._resolve_base_url()
@@ -97,9 +96,9 @@ class LoginCommand(Command):
                 time.sleep(interval)
                 continue
 
-            if (
-                response.status_code == 400
-                and detail in ("slow_down", "slow down")
+            if response.status_code == 400 and detail in (
+                "slow_down",
+                "slow down",
             ):
                 interval += 5
                 time.sleep(interval)
--- a/dotflow/cli/commands/login.py
+++ b/dotflow/cli/commands/login.py
@@ -120,6 +120,14 @@ class LoginCommand(Command):
                 time.sleep(interval)
                 continue
 
+            if (
+                response.status_code == 400
+                and detail in ("slow_down", "slow down")
+            ):
+                interval += 5
+                time.sleep(interval)
+                continue
+
             if response.status_code == 410:
                 print(settings.ERROR_ALERT, "Authorization expired.")
                 return None
--- a/dotflow/cli/commands/login.py
+++ b/dotflow/cli/commands/login.py
@@ -14,8 +14,7 @@ from dotflow.core.config_file import save_cloud_config
 from dotflow.settings import Settings as settings
 
 DEFAULT_BASE_URLS = (
-    "https://www.host.dotflow.io/api/v1",
-    "https://www.lib.dotflow.io/api/v1",
+    "https://www.cli.dotflow.io/api/v1",
 )
 DEFAULT_BASE_URL = DEFAULT_BASE_URLS[0]
 DEVICE_ENDPOINT = "/auth/cli/device"
--- a/dotflow/cli/commands/login.py
+++ b/dotflow/cli/commands/login.py
@@ -18,6 +18,7 @@ DEVICE_ENDPOINT = "/cli/device"
 TOKEN_ENDPOINT = "/cli/token"
 TIMEOUT = 15
 DEFAULT_INTERVAL = 5
+MAX_POLL_INTERVAL = 60
 
 
 class LoginCommand(Command):
@@ -100,7 +101,7 @@ class LoginCommand(Command):
                 "slow_down",
                 "slow down",
             ):
-                interval += 5
+                interval = min(interval + 5, MAX_POLL_INTERVAL)
                 time.sleep(interval)
                 continue
 
--- a/dotflow/cli/commands/login.py
+++ b/dotflow/cli/commands/login.py
@@ -25,6 +25,7 @@ DEFAULT_INTERVAL = 5
 
 
 class LoginCommand(Command):
+
     def setup(self):
         token = getattr(self.params, "token", None)
 
--- a/dotflow/cli/commands/login.py
+++ b/dotflow/cli/commands/login.py
@@ -13,10 +13,7 @@ from dotflow.cli.command import Command
 from dotflow.core.config_file import save_cloud_config
 from dotflow.settings import Settings as settings
 
-DEFAULT_BASE_URLS = (
-    "https://www.cli.dotflow.io/api/v1",
-)
-DEFAULT_BASE_URL = DEFAULT_BASE_URLS[0]
+DEFAULT_BASE_URL = "https://www.cli.dotflow.io/api/v1"
 DEVICE_ENDPOINT = "/auth/cli/device"
 TOKEN_ENDPOINT = "/auth/cli/token"
 TIMEOUT = 15
@@ -27,19 +24,16 @@ class LoginCommand(Command):
 
     def setup(self):
         token = getattr(self.params, "token", None)
+        base_url = self._resolve_base_url()
 
         if token:
-            base_url = self._explicit_base_url() or DEFAULT_BASE_URL
             save_cloud_config(token=token, base_url=base_url)
             print(settings.INFO_ALERT, "Token saved.")
             return
 
-        handshake_result = self._start_device_with_fallback()
-
-        if handshake_result is None:
+        handshake = self._start_device(base_url)
+        if handshake is None:
             return
-
-        base_url, handshake = handshake_result
 
         print(settings.INFO_ALERT, f"Opening {handshake['verification_uri']}")
         print(settings.INFO_ALERT, f"Code: {handshake['user_code']}")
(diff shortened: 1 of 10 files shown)


FernandoCelmer (commented):
🔍 Code Review

**Code issues found: 2**

| # | Severity | Comment |
|---|----------|---------|
| 1 | [Suggestion] | [Token file race: writable before chmod](https://github.com/dotflow-io/dotflow/pull/271#discussion_r3083373345) |
| 2 | [Suggestion] | [slow_down aborts login flow (RFC 8628)](https://github.com/dotflow-io/dotflow/pull/271#discussion_r3083373351) |

dotflow/core/config_file.py
```diff
@@ -0,0 +1,121 @@
...
+    content = (
+        f"[{CLOUD_SECTION}]\n"
+        f'base_url = "{_escape(base_url)}"\n'
+        f'token = "{_escape(token)}"\n'
+    )
+    path.write_text(content, encoding="utf-8")
+
+    with contextlib.suppress(OSError):
+        os.chmod(path, 0o600)
+
```
FernandoCelmer, line 70: [Suggestion]

**Problem** — `save_cloud_config()` writes the API token to disk via `path.write_text(...)` and then restricts the file to `0o600` via `os.chmod()`. Between these two operations, the file exists with the default umask permissions — typically `0o644`, which is group- and world-readable. On a shared or multi-user system, another process running as a different user could read the token during this window.

**Failure scenario** — On a shared CI/CD runner or a multi-user development server:

```
T0: dotflow login completes, calls save_cloud_config()
T1: path.write_text(content) executes
    → file created with mode 0o644 (umask default)
    → file contains: [cloud]\napi_token = "sk-live-abc123..."
T2: Malicious cron job scans ~/.dotflow/ for new files
    → reads config.toml while it's still 0o644 → token leaked!
T3: os.chmod(path, 0o600) executes
    → file now restricted to owner-only — but too late
```

The window is typically milliseconds, but on I/O-constrained systems (network filesystems, containers with slow overlayfs), it can be significantly longer.

**Fix** — Use `os.open()` with `O_CREAT | O_TRUNC` and mode `0o600` so the file is created atomically with the correct permissions. No window of vulnerability exists because the mode is applied at file creation time, before any data is written:

```python
import os

def save_cloud_config(path, content):
    path.parent.mkdir(parents=True, exist_ok=True)
    
    fd = os.open(
        str(path),
        os.O_WRONLY | os.O_CREAT | os.O_TRUNC,
        0o600,  # applied at creation time — no race window
    )
    try:
        with os.fdopen(fd, "w", encoding="utf-8") as fh:
            fh.write(content)
    except Exception:
        os.close(fd)
        raise
    # No need for os.chmod() — permissions were set at creation
```

Note: The `0o600` mode passed to `os.open()` is still subject to the process umask, but since `0o600` has no group/world bits, any umask leaves it at `0o600` or more restrictive — never less.

**References** — This is a well-known pattern in credential management. See [CWE-732: Incorrect Permission Assignment for Critical Resource](https://cwe.mitre.org/data/definitions/732.html).

dotflow/cli/commands/login.py
```diff
@@ -0,0 +1,138 @@
...
+                time.sleep(interval)
+                continue
+
+            if response.status_code == 410:
+                print(settings.ERROR_ALERT, "Authorization expired.")
+                return None
+
+            print(settings.ERROR_ALERT, f"{response.status_code}: {detail}")
+            return None
+
```
FernandoCelmer, line 114: [Suggestion]

**Problem** — The `_poll_token()` polling loop handles exactly two non-success states:
1. `400 + "authorization_pending"` → keep waiting
2. `410` → expired, abort

Any other response — including the RFC 8628 `slow_down` signal — falls through to the catch-all `print(error) + return None`, which silently aborts the entire login flow.

Per [RFC 8628 §3.5](https://www.rfc-editor.org/rfc/rfc8628#section-3.5), when a server is under load, it responds with `400 + {"error": "slow_down"}` (or `{"detail": "slow_down"}` in your API's convention), instructing the client to **increase its polling interval by 5 seconds and continue** — not abort.

**Failure scenario** — Imagine dotflow-api is under heavy load during a product launch:

```
$ dotflow login
Opening browser for authentication...
Your code: ABCD-1234

# User opens browser, starts authorizing...
# Meanwhile, server is under load

# CLI polls /auth/cli/token:
#   Response: 400 {"detail": "slow_down"}
#
# Current code falls through to:
#   print("⚠ 400: slow_down")
#   return None
#
# Login aborts! User sees:
Error: 400: slow_down

# User has to re-run `dotflow login` and re-authorize in browser
```

This is particularly frustrating because the server is explicitly telling the client "I'm busy, try again slower" — but the client interprets it as a fatal error.

**Fix** — Add a `slow_down` branch before the catch-all, following the RFC specification:

```python
if (
    response.status_code == 400
    and detail in ("slow_down", "slow down")
):
    interval += 5  # RFC 8628 §3.5: MUST increase by 5 seconds
    if settings.VERBOSE:
        print(f"  Server requested slow down, polling every {interval}s")
    time.sleep(interval)
    continue

# Existing catch-all for truly unexpected errors:
print(settings.ERROR_ALERT, f"{response.status_code}: {detail}")
return None
```

This keeps the login flow alive when the server asks for back-off, which is the intended behavior of the device authorization grant. The 5-second increment is mandated by the RFC, not a suggestion.

**References** — [RFC 8628 §3.5 — Device Authorization Grant: Token Request](https://www.rfc-editor.org/rfc/rfc8628#section-3.5)

dotflow/cli/commands/login.py
```diff
@@ -0,0 +1,146 @@
...
+                continue
+
+            if (
+                response.status_code == 400
+                and detail in ("slow_down", "slow down")
+            ):
+                interval += 5
+                time.sleep(interval)
+                continue
+
```
FernandoCelmer, line 107: [Suggestion]

**Problem** — The `slow_down` handling correctly increases the polling interval by 5 seconds per RFC 8628 §3.5, but the interval is never capped. If the server sends repeated `slow_down` responses — whether due to sustained load, a misconfigured rate limiter, or a bug — the interval grows without bound.

**Failure scenario** — The device code has `expires_in = 300` seconds (5 minutes). The server is under heavy load and sends consecutive `slow_down` responses:

```
Poll  1: interval =  10s  (initial 5 + first slow_down)
Poll  2: interval =  15s
Poll  3: interval =  20s
...
Poll 10: interval =  55s
...
Poll 15: interval =  80s
...
Poll 20: interval = 105s  ← sleeping almost 2 minutes per poll!
```

At poll 20, `time.sleep(105)` blocks the CLI for 1 minute 45 seconds. If the deadline has already passed (or passes during the sleep), the user stares at a frozen terminal with no feedback. The deadline check only runs *after* the sleep returns, so the CLI can hang well past `expires_in`.

In a worst case, if the server recovers after 30 seconds but the client is sleeping for 105 seconds, the user waits 75 seconds unnecessarily — and by then the device code may have expired server-side.

**Fix** — Cap the interval at a reasonable maximum (e.g., 30 seconds) and also clamp against the remaining deadline so the sleep never exceeds the time left:

```python
if (
    response.status_code == 400
    and detail in ("slow_down", "slow down")
):
    interval = min(interval + 5, 30)  # cap at 30s per RFC recommendation
    remaining = max(deadline - time.monotonic(), 0)
    time.sleep(min(interval, remaining))
    continue
```

This ensures:
1. The interval never grows beyond 30 seconds, regardless of how many `slow_down` responses arrive
2. The sleep never exceeds the remaining time before the deadline
3. If the deadline has passed, the loop exits immediately on the next iteration's deadline check

**References** — [RFC 8628 §3.5](https://www.rfc-editor.org/rfc/rfc8628#section-3.5) mandates increasing the interval but does not specify an upper bound. In practice, OAuth implementations commonly cap at 30s (see [Google's device flow docs](https://developers.google.com/identity/protocols/oauth2/limited-input-device)).

FernandoCelmer (commented):
🔍 Code Review (Re-review)

**Previous issues: 2/2 fixed ✓**

**New issues found: 1**

| # | Severity | Comment |
|---|----------|---------|
| 1 | [Suggestion] | [Unbounded interval growth on repeated slow_down](https://github.com/dotflow-io/dotflow/pull/271#discussion_r3083731101) |

PR #282, opened 2026-04-27
⚙️ FEATURE-#281: on_input_change policy for resume workflows

### LINKED ISSUE #281: Idempotency policy for workflow re-submission with same workflow_id but different initial_context
## Context

When integrating dotflow into a queued / job-based system (SQS → Lambda, Celery, etc.), we frequently re-trigger the same workflow with the same `workflow_id` but want different idempotency semantics depending on whether the **input** (initial_context) changed:

- **Same input, retry/resume** — dotflow already handles this beautifully via `resume=True` + persistent `Storage` provider. Steps with stored checkpoints are skipped, the workflow continues where it stopped. 
- **Different input, restart** — currently no built-in primitive. The previous run's checkpoints in `Storage` will silently be picked up by `resume=True`, which is wrong: the new input is ignored and the old `parse`/`extract` output is used.

## What we want

A native way for `DotFlow` / `Manager` to detect that `initial_context` differs from a previous run and either:

1. Throw / refuse to start (strict),
2. Invalidate prior checkpoints automatically (overwrite), or
3. Expose a hook (`on_input_change=...` callback) for the user to decide.

Today the workaround is to:
- Hash the input payload externally
- Use `workflow_id = f\"{job}-{input_hash}\"` so each input gets its own checkpoint namespace
- Manually clean up old prefixes

This works but is boilerplate every consumer reinvents.

## Proposed API sketch

```python
workflow = DotFlow(workflow_id=\"my-job\", config=config)
workflow.task.add(step=[parse, transform, load], initial_context=payload)

workflow.start(
    resume=True,
    on_input_change=\"reset\",     # \"reset\" | \"reuse\" | \"raise\"
)
```

Or more granular:

```python
workflow.start(
    resume=True,
    fingerprint=hash_of(payload),   # lib stores it next to checkpoints,
                                    # invalidates if mismatch
)
```

## Why it matters

Queued/job-based deployments are dotflow's sweet spot (Lambda + SQS, ECS, scheduled runs). Right now combining `resume=True` with any system where the input might change is footgun-prone unless you read the source carefully.

A documented + built-in idempotency story would prevent silently-wrong runs.

## Related

- `docs_src/checkpoint/checkpoint.py` shows happy path (same input, resume)
- No example shows what happens when the *same workflow_id* gets a *different initial_context*

FernandoCelmer:
Closes #281

## Description

- `dotflow/core/fingerprint.py` (new): SHA-256 fingerprint helpers (`fingerprint_of`, `fp_key`, `read_fingerprint`, `write_fingerprint`) and `VALID_POLICIES` constant for the three supported policies.
- `dotflow/core/workflow.py`: `Manager.__init__` gains `on_input_change: str = "reuse"` and calls `_enforce_input_fingerprint` on resume when storage is configured.
- `dotflow/abc/storage.py`: New non-abstract `clear(workflow_id)` method; default implementation raises `NotImplementedError`.
- `dotflow/providers/storage_default.py`: `clear()` deletes in-memory keys whose name starts with `workflow_id`.
- `dotflow/providers/storage_file.py`: `clear()` removes files in the tasks directory whose name starts with `workflow_id`.
- `dotflow/providers/storage_s3.py`: `clear()` delegates to `S3.delete_prefix(workflow_id)`.
- `dotflow/cloud/aws/services/s3.py`: `delete_prefix(sub_prefix)` paginates and batch-deletes S3 objects; rejects empty `sub_prefix` to prevent bucket-wide wipes.
- `dotflow/core/exception.py`: `InputChangedError` and `InvalidOnInputChange` exceptions with descriptive messages.
- `docs_src/checkpoint/checkpoint_input_change.py` (new): End-to-end usage example.
- `tests/core/test_fingerprint.py` (new): Unit tests for fingerprint helpers.
- `tests/core/test_workflow_input_change.py` (new): Policy integration tests (reuse / reset / raise / invalid / no-resume).
- `tests/providers/test_storage_default.py`, `test_storage_file.py`, `test_storage_s3.py`: `clear()` coverage.

## Motivation and Context

Resuming a workflow with a different `initial_context` silently reused stale checkpoints, making it impossible to detect or handle input drift. Closes #281.

## Types of changes

- [ ] Documentation

## Checklist


FernandoCelmer (commented):
placeholder

dotflow/core/fingerprint.py
```diff
@@ -0,0 +1,39 @@
...
+
+FP_KEY_SUFFIX = "-_input_fp"
+VALID_POLICIES = ("reuse", "reset", "raise")
+
+
+def fingerprint_of(values: list[Any]) -> str:
+    """SHA256 of JSON-serialized list. Falls back to repr for non-JSON inputs."""
+    try:
+        encoded = dumps(values, sort_keys=True, default=str)
+    except (TypeError, ValueError):
```
FernandoCelmer, line 25: [Blocking]

**Problem** — `json.dumps(..., default=str)` is called before the `except` branch, so any non-JSON-serializable object in `initial_context.storage` is converted via `str()`. For objects without a custom `__str__`, Python falls back to `object.__repr__`, which includes the instance's memory address (e.g. `<MyClass object at 0x7f3a1b2c4d50>`). Because the address changes every process restart, two runs with **identical** inputs produce **different** fingerprints — triggering a false mismatch on every resume.

**Failure scenario** — Suppose `initial_context` holds a Pydantic model or a dataclass that has not implemented `__str__`:

```python
from dataclasses import dataclass

@action
def step_one(initial_context):
    ...

@dataclass
class Payload:
    value: int

workflow.task.add(step=[step_one], initial_context=Payload(value=42))
workflow.start(resume=True, on_input_change="raise")
# Second run (new process, same Payload(value=42)):
# fingerprint differs → InputChangedError raised even though nothing changed
```

**Fix** — Drop `default=str` and let the `except` branch handle truly non-serializable inputs; then inside the `except` branch, use a stable representation. A simple option: use `dataclasses.asdict` / `model_dump` as a pre-processing step, or raise a clear error instructing users to pass JSON-serializable payloads. At minimum, make the fallback deterministic:

```python
def fingerprint_of(values: list[Any]) -> str:
    try:
        encoded = dumps(values, sort_keys=True)
    except (TypeError, ValueError):
        # WARNING: repr is only stable if __repr__ is deterministic.
        # Objects with memory-address-based repr will produce different
        # fingerprints across runs. Consider enforcing JSON-serializable payloads.
        encoded = repr(values)

    return sha256(encoded.encode("utf-8")).hexdigest()
```

Without `default=str`, the `except` branch is now actually reachable for non-serializable types. Add a log warning inside the except to make the instability visible to users. Alternatively, surface a `FingerprintWarning` so operators know their fingerprints may be unreliable.

**References** — [Python `json.dumps` — `default` parameter](https://docs.python.org/3/library/json.html#json.JSONEncoder)

FernandoCelmer, line 25: [Blocking]

**Problem** — `json.dumps(..., default=str)` is called before the `except` branch, so any non-JSON-serializable object in `initial_context.storage` is converted via `str()`. For objects without a custom `__str__`, Python falls back to `object.__repr__`, which includes the instance's memory address (e.g. `<MyClass object at 0x7f3a1b2c4d50>`). Because the address changes every process restart, two runs with **identical** inputs produce **different** fingerprints — triggering a false mismatch on every resume.

**Failure scenario** — Suppose `initial_context` holds a Pydantic model or a dataclass that has not implemented `__str__`:

```python
from dataclasses import dataclass

@action
def step_one(initial_context):
    ...

@dataclass
class Payload:
    value: int

workflow.task.add(step=[step_one], initial_context=Payload(value=42))
workflow.start(resume=True, on_input_change="raise")
# Second run (new process, same Payload(value=42)):
# fingerprint differs → InputChangedError raised even though nothing changed
```

**Fix** — Drop `default=str` and let the `except` branch handle truly non-serializable inputs; then inside the `except` branch, use a stable representation. A simple option: use `dataclasses.asdict` / `model_dump` as a pre-processing step, or raise a clear error instructing users to pass JSON-serializable payloads. At minimum, make the fallback deterministic:

```python
def fingerprint_of(values: list[Any]) -> str:
    try:
        encoded = dumps(values, sort_keys=True)
    except (TypeError, ValueError):
        # WARNING: repr is only stable if __repr__ is deterministic.
        # Objects with memory-address-based repr will produce different
        # fingerprints across runs. Consider enforcing JSON-serializable payloads.
        encoded = repr(values)

    return sha256(encoded.encode("utf-8")).hexdigest()
```

Without `default=str`, the `except` branch is now actually reachable for non-serializable types. Add a log warning inside the except to make the instability visible to users. Alternatively, surface a `FingerprintWarning` so operators know their fingerprints may be unreliable.

**References** — [Python `json.dumps` — `default` parameter](https://docs.python.org/3/library/json.html#json.JSONEncoder)

FernandoCelmer (commented):
placeholder

FernandoCelmer (commented):
🔍 Code Review

**Code issues found: 1**

| # | Severity | Comment |
|---|----------|---------|
| 1 | [Blocking] | [fingerprint_of: default=str produces unstable fingerprints](https://github.com/dotflow-io/dotflow/pull/282#discussion_r3149180293) |

FernandoCelmer (commented):
🔍 Code Review

**Code issues found: 4**

| # | Severity | Comment |
|---|----------|---------|
| 1 | [Blocking] | Prefix collision in `clear()` deletes sibling workflow data |
| 2 | [Blocking] | TOCTOU race in `StorageFile.clear()` |
| 3 | [Blocking] | Unstable fingerprints from `default=str` + dead `except` branch |
| 4 | [Suggestion] | `clear()` should be abstract or have a better error path |

dotflow/providers/storage_default.py
```diff
@@ -20,3 +20,10 @@ def get(self, key: str) -> Context:
 
     def key(self, task: Callable) -> str:
         return f"{task.workflow_id}-{task.task_id}"
+
```
FernandoCelmer, line 23: [Blocking]

**Problem** — `k.startswith(prefix)` uses the raw `workflow_id` as prefix, which means any workflow whose ID shares a prefix with another will have its data incorrectly deleted. For example, `clear("wf-prod")` also wipes keys belonging to `"wf-prod-v2"` since `"wf-prod-v2-task-1".startswith("wf-prod")` is `True`.

**Failure scenario** — A user runs two workflows: `pipeline` and `pipeline-v2`. When `pipeline` resumes with `on_input_change="reset"`, `clear("pipeline")` deletes all checkpoints from `pipeline-v2` as well, including its fingerprint key.

**Fix** — Anchor the prefix match to the key separator. Since all storage keys follow the format `{workflow_id}-{task_id}`, use `k.startswith(f"{workflow_id}-")` instead. This applies to all three providers (`StorageDefault`, `StorageFile`, `StorageS3`).

```python
prefix = f"{workflow_id}-"
stale = [k for k in self._store if k.startswith(prefix)]
```

The fingerprint key `{workflow_id}-_input_fp` already contains the `-` separator, so the fix is safe for all key formats.

dotflow/providers/storage_file.py
```diff
@@ -59,6 +59,13 @@ def get(self, key: str) -> Context:
     def key(self, task: Callable):
         return f"{task.workflow_id}-{task.task_id}.json"
 
+    def clear(self, workflow_id: str) -> None:
+        prefix = f"{workflow_id}"
+
```
FernandoCelmer, line 64: [Blocking]

**Problem** — `entry.unlink()` without `missing_ok=True` is a TOCTOU (time-of-check/time-of-use) race. Between `entry.is_file()` and `entry.unlink()`, another process or thread can delete the file, causing `FileNotFoundError`. This is plausible in background or parallel execution modes.

**Failure scenario** — Two concurrent workflow resets race on the same directory. Process A checks `is_file()` → True, process B deletes the file, process A calls `unlink()` → `FileNotFoundError` crashes the workflow init.

**Fix** — Use `missing_ok=True` (Python 3.8+):

```python
entry.unlink(missing_ok=True)
```

Also apply the same prefix fix as `StorageDefault`:
```python
prefix = f"{workflow_id}-"
```

dotflow/core/fingerprint.py
```diff
@@ -0,0 +1,39 @@
...
+from dotflow.core.context import Context
+
+FP_KEY_SUFFIX = "-_input_fp"
+VALID_POLICIES = ("reuse", "reset", "raise")
+
+
+def fingerprint_of(values: list[Any]) -> str:
+    """SHA256 of JSON-serialized list. Falls back to repr for non-JSON inputs."""
+    try:
+        encoded = dumps(values, sort_keys=True, default=str)
```
FernandoCelmer: [Blocking]

**Problem** — `json.dumps(..., default=str)` never raises `TypeError` — `default=str` handles all non-serializable objects. The `except` branch (`repr(values)`) is dead code that can never execute. More critically, `default=str` converts objects without a custom `__str__` into `<ClassName object at 0x7f...>`, which includes the **memory address**. This produces different fingerprints across process restarts for the same logical input.

**Failure scenario** — A user passes a dataclass or custom object as `initial_context`. Run 1 fingerprints it as `"<Payload object at 0x1234>"`, run 2 as `"<Payload object at 0x5678>"`. The fingerprints don't match, so `on_input_change="raise"` fires spuriously on every resume, and `"reset"` wipes checkpoints on every restart.

**Fix** — Remove the dead `except` branch. Document that `default=str` is used and that objects should implement a stable `__str__` or be JSON-serializable for reliable fingerprinting:

```python
def fingerprint_of(values: list[Any]) -> str:
    """SHA256 of JSON-serialized list.

    Non-serializable objects are coerced via str(). For stable
    fingerprints, ensure custom objects implement __str__ deterministically.
    """
    encoded = dumps(values, sort_keys=True, default=str)
    return sha256(encoded.encode("utf-8")).hexdigest()
```

The test `test_handles_non_json_serializable_falls_back_to_repr` should be renamed and updated to reflect this behavior — it currently asserts a hash is produced but doesn't verify stability across calls.

dotflow/abc/storage.py
```diff
@@ -23,3 +23,14 @@ def get(self, key: str) -> Context:
     @abstractmethod
     def key(self, task: Callable):
         """Function that returns a key to get and post storage"""
+
+    def clear(self, workflow_id: str) -> None:
+        """Remove every persisted entry under ``workflow_id``.
+
```
FernandoCelmer, line 30: [Suggestion]

**Problem** — `clear()` is a concrete method that raises `NotImplementedError`, but it is required for the `reset` policy to work. Any custom `Storage` subclass that forgets to implement `clear()` will get an opaque `NotImplementedError` deep inside `Manager.__init__` with no guidance about which policy triggered it.

**Fix** — Either make `clear()` an `@abstractmethod` (forcing all providers to implement it), or catch `NotImplementedError` in `_enforce_input_fingerprint` and re-raise with a descriptive message:

```python
try:
    storage.clear(workflow_id=workflow_key)
except NotImplementedError:
    raise NotImplementedError(
        f"{type(storage).__name__} does not support clear(); "
        f"on_input_change='reset' requires a storage provider that implements clear()"
    ) from None
```

PR #297, opened 2026-05-02
⚙️ FEATURE-#287: Expand Storage ABC with delete, list, TTL, CAS

### LINKED ISSUE #287: Expand `Storage` ABC with `delete_prefix`, `list_keys`, TTL, CAS
**Labels:** `enhancement`, `discovery`
**Related:** #281, #177, #178

### Context

`dotflow/abc/storage.py` exposes only three abstract methods:

```python
class Storage(ABC):
    @abstractmethod
    def post(self, key: str, context: Context) -> None: ...
    @abstractmethod
    def get(self, key: str) -> Context: ...
    @abstractmethod
    def key(self, task: Callable): ...
```

This minimal surface blocks:

- **#281 idempotency reset** — needs `delete_prefix(workflow_id)` to
invalidate stale checkpoints.
- **Run inspection / replay** — needs `list_keys(prefix)` to
enumerate a run's tasks.
- **Bounded retention** — without TTL, checkpoints accumulate
indefinitely (S3 cost, sqlite bloat).
- **Distributed coordination** — leader election or single-flight
semantics need atomic compare-and-swap (CAS).

The four current drivers (`StorageDefault`, `StorageFile`, `StorageGCS`, `StorageS3`) all natively support these primitives — the ABC is the bottleneck, not the backends.

### Concept

Extend the ABC with the missing primitives. Provide a default in-memory implementation only for `StorageDefault`; require explicit implementation for all others. Ship a contract test suite that every driver subclass must pass.

### Proposed ABC

```python
class Storage(ABC):
    @abstractmethod
    def post(
        self,
        key: str,
        context: Context,
        ttl: int | None = None,
        fingerprint: str | None = None,
    ) -> None: ...

    @abstractmethod
    def get(self, key: str) -> Context | None: ...

    @abstractmethod
    def delete(self, key: str) -> bool: ...

    @abstractmethod
    def delete_prefix(self, prefix: str) -> int:
        """Returns the number of keys removed."""

    @abstractmethod
    def list_keys(self, prefix: str) -> Iterable[str]: ...

    @abstractmethod
    def atomic_swap(self, key: str, expected, new) -> bool:
        """CAS. Returns True iff the swap happened."""

    @abstractmethod
    def key(self, task: Callable) -> str: ...
```

### Contract test suite

```python
# dotflow/testing/storage_contract.py
class StorageContract:
    """Subclass per driver and override the `storage` fixture."""
    @pytest.fixture
    def storage(self) -> Storage: raise NotImplementedError

    def test_roundtrip(self, storage): ...
    def test_delete_returns_true_when_present(self, storage): ...
    def test_delete_returns_false_when_absent(self, storage): ...
    def test_delete_prefix_returns_count(self, storage): ...
    def test_list_keys_filters_by_prefix(self, storage): ...
    def test_atomic_swap_succeeds_when_expected_matches(self, storage): ...
    def test_atomic_swap_fails_under_concurrent_write(self, storage): ...
    def test_ttl_expiration(self, storage): ...   # may skip if TTL unsupported
```

### Acceptance criteria

- [ ] ABC updated as above
- [ ] Each existing driver implements every method
- [ ] `StorageContract` test suite added
- [ ] Each driver has a `Test{Driver}Contract(StorageContract)`
subclass
- [ ] CHANGELOG entry flags this as a public ABC change

### Migration notes

External `Storage` subclasses (third-party drivers) will break. This is a public API change; ship behind a major version bump or guard with a `LegacyStorage` shim.

---

FernandoCelmer:
Closes #287

## Description

- `dotflow/abc/storage.py` — Storage ABC gains five new abstract methods: `delete`, `delete_prefix`, `list_keys`, `atomic_swap`; `post` signature extended with optional `ttl` and `fingerprint`; `clear` demoted to a concrete default that delegates to `delete_prefix`
- `dotflow/providers/storage_default.py` — Full in-memory implementation with `threading.RLock`, TTL via `time.monotonic`, lazy eviction on `get`/`list_keys`
- `dotflow/providers/storage_file.py` — File-backed implementation with sidecar `.meta` files for TTL/fingerprint; `threading.RLock` added (currently only used in `atomic_swap`)
- `dotflow/providers/storage_s3.py` — S3 implementation delegating new operations to the `S3` service class
- `dotflow/providers/storage_gcs.py` — GCS implementation delegating new operations to the `GCS` service class
- `dotflow/cloud/aws/services/s3.py` — Adds `delete` and `list_keys` to the S3 service layer
- `dotflow/cloud/gcp/services/gcs.py` — Adds `delete` and `list_keys` to the GCS service layer
- `dotflow/testing/storage_contract.py` — New reusable `StorageContract` mixin for third-party driver authors
- `dotflow/testing/__init__.py` — Exports `StorageContract`
- `tests/providers/test_storage_default_contract.py` — Runs `StorageContract` against `StorageDefault`
- `tests/providers/test_storage_file_contract.py` — Runs `StorageContract` against `StorageFile`

## Motivation and Context

The existing Storage ABC exposed only `post`, `get`, `key`, and `clear`, which was insufficient for workflows that need TTL-based expiration, fingerprint-based deduplication, enumeration of stored keys, targeted single-key deletion, and compare-and-swap updates. Closes #287.

## Types of changes

- [ ] Documentation

## Checklist

### DIFF
--- a/dotflow/cloud/aws/services/s3.py
+++ b/dotflow/cloud/aws/services/s3.py
@@ -84,9 +84,7 @@ class S3(ObjectStorage):
         except self._s3.exceptions.NoSuchKey:
             return [], None
 
-    def write_if_match(
-        self, key: str, data: list, etag: str | None
-    ) -> bool:
+    def write_if_match(self, key: str, data: list, etag: str | None) -> bool:
         """Conditional PutObject. Returns False on precondition failure."""
         from botocore.exceptions import ClientError
 
--- a/dotflow/cloud/aws/services/s3.py
+++ b/dotflow/cloud/aws/services/s3.py
@@ -46,6 +46,32 @@ class S3(ObjectStorage):
             ContentType="application/json",
         )
 
+    def delete(self, key: str) -> bool:
+        """Delete a single object."""
+        full_key = f"{self.prefix}{key}"
+
+        try:
+            self._s3.head_object(Bucket=self.bucket, Key=full_key)
+        except self._s3.exceptions.ClientError:
+            return False
+
+        self._s3.delete_object(Bucket=self.bucket, Key=full_key)
+
+        return True
+
+    def list_keys(self, sub_prefix: str) -> list[str]:
+        """Return keys starting with sub_prefix."""
+        full_prefix = f"{self.prefix}{sub_prefix}"
+        paginator = self._s3.get_paginator("list_objects_v2")
+        names = []
+        offset = len(self.prefix)
+
+        for page in paginator.paginate(Bucket=self.bucket, Prefix=full_prefix):
+            for item in page.get("Contents", []):
+                names.append(item["Key"][offset:])
+
+        return names
+
     def delete_prefix(self, sub_prefix: str) -> None:
         """Delete every object whose key starts with prefix + sub_prefix.
 
--- a/dotflow/cloud/aws/services/s3.py
+++ b/dotflow/cloud/aws/services/s3.py
@@ -7,6 +7,11 @@ from json import dumps, loads
 from dotflow.cloud.core import ObjectStorage
 from dotflow.core.exception import ModuleNotFound
 
+_PRECONDITION_CODES = {
+    "PreconditionFailed",
+    "ConditionalRequestConflict",
+}
+
 
 class S3(ObjectStorage):
     """Amazon S3 object storage."""
@@ -48,16 +53,65 @@ class S3(ObjectStorage):
 
     def delete(self, key: str) -> bool:
         """Delete a single object."""
+        from botocore.exceptions import ClientError
+
         full_key = f"{self.prefix}{key}"
 
         try:
             self._s3.head_object(Bucket=self.bucket, Key=full_key)
-        except self._s3.exceptions.ClientError:
-            return False
+        except ClientError as error:
+            code = error.response.get("Error", {}).get("Code")
+
+            if code in ("404", "NoSuchKey", "NotFound"):
+                return False
+
+            raise
 
         self._s3.delete_object(Bucket=self.bucket, Key=full_key)
 
         return True
+
+    def read_with_etag(self, key: str) -> tuple[list, str | None]:
+        """Return (data, etag). Etag is None when key is missing."""
+        try:
+            response = self._s3.get_object(
+                Bucket=self.bucket,
+                Key=f"{self.prefix}{key}",
+            )
+            data = response["Body"].read().decode("utf-8")
+
+            return loads(data), response.get("ETag")
+        except self._s3.exceptions.NoSuchKey:
+            return [], None
+
+    def write_if_match(
+        self, key: str, data: list, etag: str | None
+    ) -> bool:
+        """Conditional PutObject. Returns False on precondition failure."""
+        from botocore.exceptions import ClientError
+
+        kwargs = {
+            "Bucket": self.bucket,
+            "Key": f"{self.prefix}{key}",
+            "Body": dumps(data),
+            "ContentType": "application/json",
+        }
+
+        if etag is None:
+            kwargs["IfNoneMatch"] = "*"
+        else:
+            kwargs["IfMatch"] = etag
+
+        try:
+            self._s3.put_object(**kwargs)
+            return True
+        except ClientError as error:
+            code = error.response.get("Error", {}).get("Code")
+
+            if code in _PRECONDITION_CODES:
+                return False
+
+            raise
 
     def list_keys(self, sub_prefix: str) -> list[str]:
         """Return keys starting with sub_prefix."""
--- a/dotflow/providers/storage_default.py
+++ b/dotflow/providers/storage_default.py
@@ -71,7 +71,14 @@ class StorageDefault(Storage):
 
             return [k for k in self._store if k.startswith(prefix)]
 
-    def atomic_swap(self, key: str, expected: Any, new: Any) -> bool:
+    def atomic_swap(
+        self,
+        key: str,
+        expected: Any,
+        new: Any,
+        ttl: int | None = None,
+        fingerprint: str | None = None,
+    ) -> bool:
         with self._lock:
             current = self._store.get(key)
             current_value = (
@@ -83,6 +90,14 @@ class StorageDefault(Storage):
 
             payload = new if isinstance(new, Context) else Context(storage=new)
             self._store[key] = payload
+            self._fingerprints.pop(key, None)
+            self._expirations.pop(key, None)
+
+            if fingerprint is not None:
+                self._fingerprints[key] = fingerprint
+
+            if ttl is not None:
+                self._expirations[key] = time.monotonic() + ttl
 
             return True
 
--- a/dotflow/providers/storage_default.py
+++ b/dotflow/providers/storage_default.py
@@ -1,29 +1,103 @@
 """Storage Default"""
 
-from collections.abc import Callable
+from __future__ import annotations
+
+import threading
+import time
+from collections.abc import Callable, Iterable
+from typing import Any
 
 from dotflow.abc.storage import Storage
 from dotflow.core.context import Context
 
 
 class StorageDefault(Storage):
-    """In-memory storage using a dictionary."""
+    """In-memory storage."""
 
     def __init__(self):
         self._store: dict[str, Context] = {}
+        self._fingerprints: dict[str, str] = {}
+        self._expirations: dict[str, float] = {}
+        self._lock = threading.RLock()
 
-    def post(self, key: str, context: Context) -> None:
-        self._store[key] = context
+    def post(
+        self,
+        key: str,
+        context: Context,
+        ttl: int | None = None,
+        fingerprint: str | None = None,
+    ) -> None:
+        with self._lock:
+            self._store[key] = context
+
+            if fingerprint is not None:
+                self._fingerprints[key] = fingerprint
+
+            if ttl is not None:
+                self._expirations[key] = time.monotonic() + ttl
+            else:
+                self._expirations.pop(key, None)
 
     def get(self, key: str) -> Context:
-        return self._store.get(key, Context())
+        with self._lock:
+            self._evict_if_expired(key)
+
+            return self._store.get(key, Context())
+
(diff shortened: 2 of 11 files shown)


dotflow/cloud/aws/services/s3.py
```diff
@@ -46,6 +46,32 @@ def write(self, key: str, data: list) -> None:
...
         )
 
+    def delete(self, key: str) -> bool:
+        """Delete a single object."""
+        full_key = f"{self.prefix}{key}"
+
+        try:
+            self._s3.head_object(Bucket=self.bucket, Key=full_key)
+        except self._s3.exceptions.ClientError:
+            return False
```
FernandoCelmer: [Blocking]

**Problem** — `self._s3.exceptions.ClientError` does not exist on a boto3 service client. The `client.exceptions` namespace only exposes *service-model-defined* exceptions (e.g. `NoSuchBucket`, `NoSuchKey`). `ClientError` lives in `botocore.exceptions`, not in the boto3 client exceptions factory. When `head_object` receives a 404, Python evaluates the `except` expression, `self._s3.exceptions.ClientError` raises `AttributeError`, and that error propagates — the `return False` branch never executes.

The existing `read()` method in this same file correctly uses `self._s3.exceptions.NoSuchKey`, which *is* a service-model exception for `get_object`. `head_object` has no modeled 404 exception — it always raises `botocore.exceptions.ClientError` with error code `"404"`.

**Failure scenario** — `StorageS3.delete(key="absent-key")` raises `AttributeError: ClientError` instead of returning `False`. Every caller that depends on the boolean result — including the `StorageContract` suite once it runs against a real bucket — crashes.

**Fix**:
```python
from botocore.exceptions import ClientError

def delete(self, key: str) -> bool:
    full_key = f"{self.prefix}{key}"
    try:
        self._s3.head_object(Bucket=self.bucket, Key=full_key)
    except ClientError as e:
        if e.response["Error"]["Code"] in ("404", "NoSuchKey"):
            return False
        raise
    self._s3.delete_object(Bucket=self.bucket, Key=full_key)
    return True
```

**References** — [boto3 error handling guide](https://boto3.amazonaws.com/v1/documentation/api/latest/guide/error-handling.html)

dotflow/providers/storage_default.py
```diff
@@ -1,29 +1,103 @@
...
+            current_value = (
+                current.storage if isinstance(current, Context) else current
+            )
+
+            if current_value != expected:
+                return False
+
+            payload = new if isinstance(new, Context) else Context(storage=new)
+            self._store[key] = payload
+
```
FernandoCelmer, line 101: [Blocking]

**Problem** — `atomic_swap` writes the new `payload` to `self._store[key]` but does not touch `self._expirations`. If the original value was stored with a TTL, the new value inherits that expiration silently. A successful CAS can produce a value that expires within milliseconds if the original TTL was nearly exhausted.

**Failure scenario** — `post("k", ctx, ttl=10)`. Nine seconds later `atomic_swap("k", expected=old, new=fresh)` returns `True`. One second later `get("k")` calls `_evict_if_expired` and removes the freshly-swapped value because `_expirations["k"]` still points to the original deadline. The caller observes a successful swap followed by a silent, invisible deletion.

**Fix** — Clear the expiration (and fingerprint) after a successful swap:
```python
payload = new if isinstance(new, Context) else Context(storage=new)
self._store[key] = payload
self._expirations.pop(key, None)   # remove inherited TTL
self._fingerprints.pop(key, None)  # remove inherited fingerprint
return True
```
If the intended behaviour is to preserve the TTL across a swap, that decision should be documented explicitly in both this method and the ABC docstring.

dotflow/providers/storage_file.py
```diff
@@ -51,20 +79,102 @@ def get(self, key: str) -> Context:
...
+                current.storage if isinstance(current, Context) else None
+            )
+
+            if current_value != expected:
+                return False
+
+            self.delete(key)
+            payload = new if isinstance(new, Context) else Context(storage=new)
+            self.post(key=key, context=payload)
+
```
FernandoCelmer, line 167: [Blocking]

**Problem** — `atomic_swap` calls `self.delete(key)` (which removes both the data file and `{key}.meta`) and then calls `self.post(key=key, context=payload)` with no `ttl` or `fingerprint` arguments. Inside `post`, `meta` is empty so `_write_meta` is never called. This is correct when the original key had no TTL. However, if the original key *did* have a TTL, `delete` removed the old `.meta`, `post` writes no new `.meta`, and the new value has no expiration — which is the *correct* outcome. The actual latent bug is the inverse: **if a new TTL or fingerprint should be applied to the swapped-in value, there is no way to pass it through `atomic_swap`**. The ABC signature `atomic_swap(key, expected, new)` has no `ttl` or `fingerprint` parameters, so callers cannot set TTL on the result of a CAS. This is an API completeness gap.

**Failure scenario** — A caller uses `atomic_swap` to rotate a short-lived token. The new value should also have a TTL, but `atomic_swap` provides no way to express this, so the swapped value is stored without expiration.

**Fix** — Add optional `ttl` and `fingerprint` parameters to `atomic_swap` in the ABC and all driver implementations:
```python
def atomic_swap(self, key, expected, new, *, ttl=None, fingerprint=None) -> bool:
    ...
    self.post(key=key, context=payload, ttl=ttl, fingerprint=fingerprint)
```

dotflow/providers/storage_file.py
```diff
@@ -12,17 +16,25 @@
...
-    """Storage"""
+    """File-backed storage."""
 
     def __init__(self, *args, path: str = settings.START_PATH, **kwargs):
         self.path = Path(path, "tasks")
         self.path.mkdir(parents=True, exist_ok=True)
-
-    def post(self, key: str, context: Context) -> None:
+        self._lock = threading.RLock()
+
```
FernandoCelmer, line 25: [Blocking]

**Problem** — `self._lock = threading.RLock()` is initialised in `__init__`, signalling intent for thread safety, but only `atomic_swap` acquires it. `post`, `get`, `delete`, `delete_prefix`, and `list_keys` all operate on the filesystem without holding the lock. Two concurrent `post` calls for the same key share a read-modify-write cycle that is not serialised.

**Failure scenario** — Two workflow tasks finish simultaneously and both call `post("wf-1-task-A.json", ...)`. Thread A reads the existing file (empty list). Thread B reads the file (still empty). Thread A appends its context and writes `[ctx_A]`. Thread B appends its context and writes `[ctx_B]`. The final file contains only `[ctx_B]`; Thread A's result is lost with no exception raised.

**Fix** — Acquire `self._lock` in every public method:
```python
def post(self, key, context, ttl=None, fingerprint=None):
    with self._lock:
        ...  # existing body

def get(self, key):
    with self._lock:
        ...
```
Because the lock is already an `RLock`, the reentrant calls inside `atomic_swap` (which calls `get`, `delete`, and `post` while holding the lock) continue to work without deadlock.

dotflow/providers/storage_s3.py
```diff
@@ -70,16 +81,52 @@ def get(self, key: str) -> Context:
             return self._loads(storage=task_context[0])
 
         contexts = Context(storage=[])
+
         for context in task_context:
             contexts.storage.append(self._loads(storage=context))
 
         return contexts
```
FernandoCelmer, line 89: [Suggestion]

**Problem** — `atomic_swap` follows a read → compare → delete → write sequence with no distributed locking or conditional write. Two concurrent callers that both read the same current value will both pass the comparison check and both complete the swap. The method is named `atomic_swap` and the ABC declares it as `"Replace value when current equals expected"`, which implies exactly-once CAS semantics.

**Failure scenario** — Two processes call `atomic_swap("k", expected="v1", new="v2")` concurrently. Both read `"v1"`, both pass the equality check, both delete the key, and both write `"v2"`. Both return `True` — a second independent "CAS" succeeded after the first already mutated the key, defeating any optimistic-concurrency use case (deduplication guard, distributed counter, etc.).

**Fix** — At minimum, document the limitation clearly in the method and in the ABC. A real fix for S3 uses the `If-None-Match` / `If-Match` conditional PutObject header (available in S3 since 2024) or an external lock. For GCS, use object generation preconditions.

```python
def atomic_swap(self, key: str, expected: Any, new: Any) -> bool:
    """Best-effort CAS. Not atomic under concurrent writers.
    Use a distributed lock for strict once-only semantics.
    """
```

**References** — [S3 conditional writes](https://aws.amazon.com/about-aws/whats-new/2024/11/amazon-s3-functionality-conditional-writes/), [GCS object preconditions](https://cloud.google.com/storage/docs/request-preconditions)

FernandoCelmer (commented):
🔍 Code Review

**Code issues found: 5**

| # | Severity | Comment |
|---|----------|---------|
| 1 | [Blocking] | [S3.delete catches non-existent exception type](https://github.com/dotflow-io/dotflow/pull/297#discussion_r3177347666) |
| 2 | [Blocking] | [atomic_swap inherits stale TTL in StorageDefault](https://github.com/dotflow-io/dotflow/pull/297#discussion_r3177347668) |
| 3 | [Blocking] | [atomic_swap cannot propagate TTL in StorageFile](https://github.com/dotflow-io/dotflow/pull/297#discussion_r3177347669) |
| 4 | [Blocking] | [StorageFile lock only protects atomic_swap](https://github.com/dotflow-io/dotflow/pull/297#discussion_r3177347670) |
| 5 | [Suggestion] | [S3/GCS atomic_swap is not truly atomic](https://github.com/dotflow-io/dotflow/pull/297#discussion_r3177347672) |

FernandoCelmer (commented):
Code Review (re-review after fixes)

**All 5 issues from previous review have been resolved:**

1. S3 `delete` — now uses `from botocore.exceptions import ClientError` with proper 404 code filtering
2. `StorageDefault.atomic_swap` — clears old TTL/fingerprint before writing, applies new TTL if provided
3. `StorageFile.atomic_swap` — entire swap operation now within single `with self._lock:` block
4. `StorageFile` lock coverage — `post()` and `get()` now acquire `self._lock`
5. S3/GCS `atomic_swap` — now uses conditional writes (S3 `IfMatch`/`IfNoneMatch` ETags, GCS `if_generation_match`). True CAS semantics.

**Verdict:** All blockers resolved. Ready to merge.

PR #299, opened 2026-05-03
⚙️ FEATURE-#290: Decouple Task status side effects via EventBus

### LINKED ISSUE #290: Internal event bus to decouple `Task` status from logging,
notifications, and metrics

**Labels:** `enhancement`

### Context

`dotflow/core/task.py:251-267` runs three side effects inside the `status` setter:

```python
@status.setter
def status(self, value):
    self._status = value
    self.config.notify.hook_status_task(task=self)
    if value == FAILED:
        self.config.log.error(task=self)
        self.config.metrics.task_failed(task=self)
    elif value == RETRY:
        ...
```

Problems:

- A property assignment performs HTTP (notify), I/O (log), and a
counter update (metrics).
- Tests that flip status need three mocks every time.
- Adding a new observer (Sentry tracer, audit log, custom webhook)
means editing `Task` itself.
- Failure ordering is undocumented: notify runs first; if it raises,
log and metric never execute.

### Concept

Introduce a tiny in-process event bus. The setter mutates state and emits a `StatusChanged` event. Three default subscribers (`LogSubscriber`, `NotifySubscriber`, `MetricsSubscriber`) each handle the event independently and are registered in `Config`.

### Proposed implementation

```python
# dotflow/core/events.py
from collections.abc import Callable
from dataclasses import dataclass
from typing import Any
from dotflow.logging import logger

@dataclass
class StatusChanged:
    task: "Task"
    old: "TypeStatus"
    new: "TypeStatus"

class EventBus:
    def __init__(self) -> None:
        self._subs: list[Callable[[Any], None]] = []

    def subscribe(self, handler: Callable[[Any], None]) -> None:
        self._subs.append(handler)

    def emit(self, event: Any) -> None:
        for handler in self._subs:
            try:
                handler(event)
            except Exception:
                logger.exception("event handler failed")

# dotflow/core/task.py
@status.setter
def status(self, value):
    old = self._status
    self._status = value
    self.config.events.emit(StatusChanged(task=self, old=old, new=value))

# dotflow/core/config.py
from dotflow.core.events import EventBus
from dotflow.core.subscribers import (
    LogSubscriber, NotifySubscriber, MetricsSubscriber,
)

class Config:
    def __init__(self, ...):
        ...
        self.events = EventBus()
        self.events.subscribe(LogSubscriber(self.log))
        self.events.subscribe(NotifySubscriber(self.notify))
        self.events.subscribe(MetricsSubscriber(self.metrics))
```

A subscriber example:

```python
class LogSubscriber:
    def __init__(self, log): self.log = log

    def __call__(self, event):
        if not isinstance(event, StatusChanged):
            return
        if event.new == TypeStatus.FAILED:
            self.log.error(task=event.task)
        elif event.new == TypeStatus.RETRY:
            self.log.warning(task=event.task)
        else:
            self.log.info(task=event.task)
```

### Benefits

- Failure in one subscriber no longer cascades.
- Adding a new observer = `events.subscribe(my_handler)`. No edits
to `Task`.
- The bus naturally supports new event types
(`CheckpointSaved`, `WorkflowStarted`) without re-touching `Task`.

### Acceptance criteria

- [ ] `EventBus`, `StatusChanged` defined in
`dotflow/core/events.py`
- [ ] `LogSubscriber`, `NotifySubscriber`, `MetricsSubscriber`
defined in `dotflow/core/subscribers.py`
- [ ] `Task.status` setter emits the event and does no other I/O
- [ ] `Config` wires the three default subscribers
- [ ] Existing observable behavior preserved (verified by current
tests)
- [ ] New test: a raising subscriber does not stop other subscribers

### Out of scope

Async subscribers (handler runs in a worker thread). The synchronous bus is enough for the current observability surface.

---

FernandoCelmer:
## Description

- `dotflow/core/events.py` (new): `EventBus` class and `StatusChanged` dataclass for in-process pub/sub
- `dotflow/core/subscribers.py` (new): `LogSubscriber`, `NotifySubscriber`, `MetricsSubscriber` — each routes `StatusChanged` events to the corresponding provider
- `dotflow/core/config.py`: `Config.__init__` creates an `EventBus` instance and wires the three default subscribers after provider validation
- `dotflow/core/task.py`: `Task.status` setter now captures the previous status and emits `StatusChanged` instead of calling providers directly
- `tests/core/test_events.py` (new): 4 tests covering `EventBus` emit, subscriber isolation, and ordering guarantees
- `tests/core/test_subscribers.py` (new): 10 tests covering all status branches for each subscriber

## Motivation and Context

Task status side effects (logging, notification, metrics) were hardcoded directly in the `Task.status` setter, coupling the task model to concrete provider calls. This change decouples the model from its observers via an event bus, making it straightforward to add or replace subscribers without modifying `Task` or `Config`.

Closes #290

## Types of changes

- [ ] Documentation

## Checklist

### DIFF
--- a/dotflow/core/events.py
+++ b/dotflow/core/events.py
@@ -1,0 +1,39 @@
+"""Event bus"""
+
+from __future__ import annotations
+
+from collections.abc import Callable
+from dataclasses import dataclass
+from typing import TYPE_CHECKING, Any
+
+from dotflow.logging import logger
+
+if TYPE_CHECKING:
+    from dotflow.core.task import Task
+    from dotflow.core.types.status import TypeStatus
+
+
+@dataclass
+class StatusChanged:
+    """Emitted when a task transitions to a new status."""
+
+    task: Task
+    old: TypeStatus
+    new: TypeStatus
+
+
+class EventBus:
+    """In-process pub/sub. Subscribers run in registration order."""
+
+    def __init__(self) -> None:
+        self._subs: list[Callable[[Any], None]] = []
+
+    def subscribe(self, handler: Callable[[Any], None]) -> None:
+        self._subs.append(handler)
+
+    def emit(self, event: Any) -> None:
+        for handler in self._subs:
+            try:
+                handler(event)
+            except Exception:
+                logger.exception("event handler failed")
--- a/dotflow/core/config.py
+++ b/dotflow/core/config.py
@@ -9,7 +9,13 @@ from dotflow.abc.scheduler import Scheduler
 from dotflow.abc.server import Server
 from dotflow.abc.storage import Storage
 from dotflow.abc.tracer import Tracer
+from dotflow.core.events import EventBus
 from dotflow.core.exception import NotCallableObject
+from dotflow.core.subscribers import (
+    LogSubscriber,
+    MetricsSubscriber,
+    NotifySubscriber,
+)
 from dotflow.providers.log_default import LogDefault
 from dotflow.providers.metrics_default import MetricsDefault
 from dotflow.providers.notify_default import NotifyDefault
@@ -90,6 +96,11 @@ class Config:
         self.metrics = metrics if metrics is not None else MetricsDefault()
 
         self._validate()
+
+        self.events = EventBus()
+        self.events.subscribe(LogSubscriber(self.log))
+        self.events.subscribe(NotifySubscriber(self.notify))
+        self.events.subscribe(MetricsSubscriber(self.metrics))
 
     def _validate(self) -> None:
         for name, abc in self._PROVIDERS.items():
--- a/dotflow/core/subscribers.py
+++ b/dotflow/core/subscribers.py
@@ -1,0 +1,62 @@
+"""Default subscribers for the internal event bus."""
+
+from __future__ import annotations
+
+from typing import TYPE_CHECKING
+
+from dotflow.core.events import StatusChanged
+from dotflow.core.types.status import TypeStatus
+
+if TYPE_CHECKING:
+    from dotflow.abc.log import Log
+    from dotflow.abc.metrics import Metrics
+    from dotflow.abc.notify import Notify
+
+
+class NotifySubscriber:
+    """Forwards status changes to the configured notify provider."""
+
+    def __init__(self, notify: Notify):
+        self.notify = notify
+
+    def __call__(self, event):
+        if not isinstance(event, StatusChanged):
+            return
+
+        self.notify.hook_status_task(task=event.task)
+
+
+class LogSubscriber:
+    """Routes status changes to the configured log provider."""
+
+    def __init__(self, log: Log):
+        self.log = log
+
+    def __call__(self, event):
+        if not isinstance(event, StatusChanged):
+            return
+
+        if event.new == TypeStatus.FAILED:
+            self.log.error(task=event.task)
+        elif event.new == TypeStatus.RETRY:
+            self.log.warning(task=event.task)
+        else:
+            self.log.info(task=event.task)
+
+
+class MetricsSubscriber:
+    """Updates task counters on status transitions."""
+
+    def __init__(self, metrics: Metrics):
+        self.metrics = metrics
+
+    def __call__(self, event):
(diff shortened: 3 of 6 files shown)


dotflow/core/events.py
```diff
@@ -0,0 +1,39 @@
...
+    from dotflow.core.types.status import TypeStatus
+
+
+@dataclass
+class StatusChanged:
+    """Emitted when a task transitions to a new status."""
+
+    task: Task
+    old: TypeStatus
+    new: TypeStatus
```
FernandoCelmer, line 22: [Suggestion]

**Problem** — The `old` field is typed as `TypeStatus`, but the very first status transition on every `Task` emits `StatusChanged(old=None, ...)` because `Task._status` is initialised to `None` in `TaskInstance.__init__` and the status setter captures `old = self._status` before any value has been set.

The type annotation creates a false contract: anyone writing a custom subscriber who reads `event.old` and treats it as a guaranteed `TypeStatus` will get an `AttributeError` or silent `None`-comparison failure on the very first event.

**Failure scenario**

```python
class AuditSubscriber:
    def __call__(self, event):
        if isinstance(event, StatusChanged):
            # Crashes with AttributeError on first task creation:
            # 'NoneType' object has no attribute 'value'
            print(f"Transition: {event.old.value} → {event.new.value}")
```

Every `Task(...)` call immediately triggers this path (`self.status = TypeStatus.NOT_STARTED` in `__init__`), so the crash happens on construction, not during execution.

**Fix** — Narrow the annotation to reflect reality:

```python
@dataclass
class StatusChanged:
    task: Task
    old: TypeStatus | None   # None on the very first transition
    new: TypeStatus
```

Alternatively, initialise `Task._status` to `TypeStatus.NOT_STARTED` in `TaskInstance.__init__` so the field is never `None`, which would let the annotation stay as `TypeStatus` and also remove the `if not self._status` guard in the `status` getter.

**References** — [PEP 484 — Type Hints](https://peps.python.org/pep-0484/#using-none)

dotflow/core/events.py
```diff
@@ -0,0 +1,39 @@
...
+    """In-process pub/sub. Subscribers run in registration order."""
+
+    def __init__(self) -> None:
+        self._subs: list[Callable[[Any], None]] = []
+
+    def subscribe(self, handler: Callable[[Any], None]) -> None:
+        self._subs.append(handler)
+
+    def emit(self, event: Any) -> None:
+        for handler in self._subs:
```
FernandoCelmer, line 35: [Suggestion]

**Problem** — `emit` passes `event` with a live reference to the `Task` object inside `StatusChanged`. Nothing prevents a subscriber from mutating `task.status` — which calls `emit` again — producing unbounded recursion that exhausts the call stack.

The three built-in subscribers (`LogSubscriber`, `NotifySubscriber`, `MetricsSubscriber`) are safe today, but the pattern is invisible to users implementing custom subscribers. The docstring says *"Subscribers run in registration order"* but says nothing about re-entrancy.

**Failure scenario**

```python
class AutoRetrySubscriber:
    def __call__(self, event):
        if isinstance(event, StatusChanged):
            if event.new == TypeStatus.FAILED:
                # Attempting a retry — sets status → triggers emit → calls this again
                event.task.status = TypeStatus.RETRY  # RecursionError
```

**Fix** — Add a simple re-entrancy guard:

```python
class EventBus:
    def __init__(self) -> None:
        self._subs: list[Callable[[Any], None]] = []
        self._emitting: bool = False

    def emit(self, event: Any) -> None:
        if self._emitting:
            return   # or raise, or queue for later — depends on desired semantics
        self._emitting = True
        try:
            for handler in self._subs:
                try:
                    handler(event)
                except Exception:
                    logger.exception("event handler failed")
        finally:
            self._emitting = False
```

Document the constraint clearly in the class docstring so that subscriber authors know status mutations inside a handler are not allowed (or are deferred).

**References** — [Observer pattern — Re-entrancy considerations](https://refactoring.guru/design-patterns/observer)

FernandoCelmer (commented):
Code review — 2 issues found.

